PR #104, opened 2015-04-09
fix #76 - viewer does not open with multi-segment seds

### LINKED ISSUE #76: If initially closed, the SED Viewer does not open for multi-segment SEDs
Workaround: select a single-segment SED first, open the Viewer, then switch to the multi-segment SED.

New (02/25): On older versions of Iris, if the SED Viewer was closed, and then you switch to another populated (at least 1 segment) SED, the Viewer opens up automatically. This no longer happens on the v2.1 build. If I close the Viewer and switch between two 1-segment SEDs, the Viewer does NOT automatically popup; I need to open it myself by clicking on the Viewer icon.

olaurino:
Proposed bug fix for #76 

@jbudynk please test this bug fix


jbudynk:
The bug is mostly fixed. I'm able to open the Viewer when a multi-segment SED is highlighted and I click on the SED Viewer desktop icon. However, like reported in #76, I'm still unable to reopen the Viewer by just clicking on a SED in the Builder:

> On older versions of Iris, if the SED Viewer was closed, and then you switch to another populated (at least 1 segment) SED, the Viewer opens up automatically. This no longer happens on the v2.1 build. If I close the Viewer and switch between two 1-segment SEDs, the Viewer does NOT automatically popup; I need to open it myself by clicking on the Viewer icon.

olaurino:
@jbudynk I believe the reported behavior is the correct one: if the user closes the visualizer Iris assumes the user does not want the visualizer to pop up on its own. As long as clicking on the Visualizer's icon makes the visualizer pop up and switching SEDs in one component (e.g. the Builder) results in the Visualizer to update itself, I would say the program is working as expected.

Or am I missing something?

However, there is still something wrong: If I add a new empty SED and I click on the visualizer icon, the "old" SED is displayed. A message about the SED being empty should be displayed instead.

I get this behavior in the current release/2.1 branch, though.

Commit fa189af0 shows the correct behavior.

olaurino:
@jbudynk could you please test the new fixes?

- Viewer should now correctly reopen itself when a new sed is selected (and the viewer itself was previously closed).
- Viewer should now correctly close itself when a sed is displayed and an empty sed is selected from a different component.
- Viewer should now show an empty window when the Viewer icon is clicked while an empty sed is selected.

Do you think we should print a warning message instead of displaying an empty window?

jbudynk:
On commit 375c7fe0502

1. Not working. The Viewer isn't reopening when a new SED is selected AND the Viewer was previously closed. However, **I think this makes sense** - If the user closed the Viewer, they don't want it popping-up every time they switch to a populated SED.
2. Works.
3. Works.

Hmmm.... I think I'd make a few changes to the Viewer behavior for empty SEDs. Here are the behaviors I was not expecting:

1. If I move the Viewer somewhere else on the desktop, switch to an empty SED, and back to a populated SED, the Viewer is no longer in the position I put it in before - it re-centers to the default location. I'd expect the Viewer to return to the same position it was in before.
2. I don't like the "empty" window that pops up when I open the Viewer on an empty SED. I was thinking the window should be the Viewer (with the axes, labels, the works), but without data points. **OR**, instead of displaying the IRIS logo, have a message that says "No data" or "Sed.1 is empty (no data)" or something like that. In this case, I would make the size of the window the same as the default Viewer size for a populated SED.
3. I'm also unsure if the Viewer should close or not when a user selects an empty SED... at first, I'd expect that that whatever message I get from point (2) above should happen here. But after playing around with it, hiding the Viewer for empty SEDs makes sense... I could go either way with this one.

olaurino:
@jbudynk Older commits (e.g. 9f4b0d9) show the correct viewer behavior: for instance the viewer panel is only hidden when closed by the user, so when it gets opened again it keeps its original position. However, we had to roll back those changes (see 9451914) because when the user tries to close the fitting manager panel without losing the model, the viewer is broken.

I'll see if I can combine changes in those different commits, but I suspect we will need to give up on something of the requirements, as we cannot access the third party component we use for visualization.

olaurino:
I replaced the previous commits with a new one.

My tests seems to be OK. The viewer window disappears when an empty SED is displayed. When the viewer is opened again, the location is reset. Also, the viewer will open itself when switching to a non-empty SED.

Other than that, switching among SEDs should now be OK. The viewer should correctly display multi-segment SEDs, be closed and opened normally. When closed, clicking on its icon on the desktop results in the correct SED to be displayed, unless the SED is empty, in which case the viewer will be empty, too.

This fix is not optimal, but if fixes the issue in subject, and it does not seem to introduce regressions.

I also tested the interactions with the fitting tool: after closing the fitting tool frame, the viewer displays the correct SED, according to my test.

@jbudynk please test on your side. As this is a pesky issue, tests should be thorough.

olaurino:
Editing SEDs results in the fitting manager to complain and ask the user to confirm they want to lose the model. This must be dealt with outside of the specview component, as the event that causes the fitting manager to destroy the model is Sed.CHANGED, which is fired after the SED is indeed changed. Thus if, at this point, the user clicks on "Cancel", the model is not thrown away, but now the SED in the builder and the SED in the fitting manager have diverged. The fitting manager requires that a SED is re-ingested into specview's internal format, but our attempts to force this were unfruitful and actually introduces new regressions. As we have limited control over the viewer and fitter components, the only possible (simple) solution seems to be to warn the user and let them know they must close the fitting manager window when they edit an SED. Commit d93ada81 does this.

Updates against new_devel are about to be force-pushed to this branch, but with the new_devel base.

For the record, this PR previously had commit 570aee8.

This PR should now be closed and re-opened against the new_devel branch.

Issue #55, opened 2014-11-14
Fitting tool does not react when Sed is null


olaurino:

olaurino:
Unfortunately, this bug was added back when trying to fix other VIewer-related bugs.

olaurino:
This should have been fixed with the history review: in the new history the regression never occurred.

PR #108, opened 2015-04-10
fix issue #55 that was introduced back by recent commits

olaurino:
This PR fixes issue #55 which was introduced back by commit 9451914

### DIFF
--- a/iris-specview/src/main/java/spv/components/IrisVisualizer.java
+++ b/iris-specview/src/main/java/spv/components/IrisVisualizer.java
@@ -460,6 +460,10 @@ public class IrisVisualizer implements IrisComponent {
 
                         try {
                             Spectrum sp = factory.readAllSegments(null, sed);
+                            if (sp == null) {
+                                NarrowOptionPane.showMessageDialog(ws.getRootFrame(), "No SEDs to fit. Please load a file.", "Fitting Engine", NarrowOptionPane.INFORMATION_MESSAGE);
+                                return;
+                            }
                             sp.setName(sed.getId());
 
                             // Get model manager from Sed attachment


jbudynk:
This works. However, there were two unexpected behaviors I noticed:

1. If I remove a segment from the SED when it's in a fitting session, the Fitting Tool asks me to cancel the session of not. If I choose "Cancel," I don't see the red line representing the fit anymore. The fitting tool is stuck on the fitting the old data before I removed the segment. I know I'm supposed to close the session, but if a user saw this message, they might be confused and say "No, I want to keep fitting my SED, but without that segment there," and keep the fitting session open.

   Instead of bringing up the default messages for removing a segment and  closing a fitting session, is it possible to check if there's a SED has a fitting session attached to it, so that we can tell the user "If you remove a segment during a fitting session, the fitting session will close"?

2. If I have three SEDs opened in the Builder -- Sed.1 populated, Sed.2 populated, and Sed.3 empty -- and I open a fitting session on Sed.1, then switch to Sed.3 in the Builder, the Fitting Tool windows don't disappear. If I switch from Sed.1 to Sed.2, the Fitting Tool windows disappear as expected.

olaurino:
Editing SEDs results in the fitting manager to complain and ask the user to confirm they want to lose the model. This must be dealt with outside of the specview component, as the event that causes the fitting manager to destroy the model is Sed.CHANGED, which is fired after the SED is indeed changed. Thus if, at this point, the user clicks on "Cancel", the model is not thrown away, but now the SED in the builder and the SED in the fitting manager have diverged. The fitting manager requires that a SED is re-ingested into specview's internal format, but our attempts to force this were unfruitful and actually introduces new regressions. As we have limited control over the viewer and fitter components, the only possible (simple) solution seems to be to warn the user and let them know they *must* close the fitting manager window when they edit an SED. Commit d93ada81 does this.

Updates against new_devel are about to be force-pushed to this branch, but with the new_devel base.

For the record, this PR previously had commit 570aee8.

This PR should now be closed and re-opened against the new_devel branch.

PR #113, opened 2015-04-21
fix #107 - fix flux results when integrating model

### LINKED ISSUE #107: Fix flux results when integrating under a model
Flux values from integrating under a model are incorrect. Increasing the passband range high end-point results in lower flux values, which is incorrect - the flux should get higher. This is because we divide the resultant flux by the effective wavelength to deal with units.

Instead, we should convert the model to units erg/s/cm2/Angstrom vs Angstrom. In this way, we don't need to divide by the effective wavelength to get the flux in valid Specview units; the flux results should already be in erg/s/cm2

olaurino:
fix #107 - fix flux results when integrating model

@jbudynk please test.


olaurino:
Closing and replacing with a new PR against new_develop

PR #116, opened 2015-04-24
Bug 76 multi segment viewer

olaurino:
This PR replaces #104.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/sed/ExtSed.java
+++ b/iris-common/src/main/java/cfa/vo/iris/sed/ExtSed.java
@@ -45,6 +45,7 @@ import spv.util.YUnits;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Method;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


olaurino:
I am copying the last comment of #104 for convenience:

Editing SEDs results in the fitting manager to complain and ask the user to confirm they want to lose the model. This must be dealt with outside of the specview component, as the event that causes the fitting manager to destroy the model is Sed.CHANGED, which is fired after the SED is indeed changed. Thus if, at this point, the user clicks on "Cancel", the model is not thrown away, but now the SED in the builder and the SED in the fitting manager have diverged. The fitting manager requires that a SED is re-ingested into specview's internal format, but our attempts to force this were unfruitful and actually introduces new regressions. As we have limited control over the viewer and fitter components, the only possible (simple) solution seems to be to warn the user and let them know they must close the fitting manager window when they edit an SED. Commit d93ada8 does this.

Updates against new_devel have been force-pushed to this branch, but with the new_devel base.

PR #162, opened 2015-10-22
140 setup travis ci

olaurino:
[release-note]
N/A
[release-note]

[story] #140 [/story]

# Description
This PR introduces a basic Travis CI configuration for Iris.
The `maven` repository as well as the `miniconda` installation are cached for efficiency.
xfvb is configured although it should not really be used at this point.

# Testing
The following Travis CI build shows that Travis builds are active and the Iris tests pass:
https://travis-ci.org/ChandraCXC/iris/builds/86906266

Issue #142, opened 2015-10-21
Headless Configuration


eholum:
Unit/functional configuration adjusted in POMs

PR #163, opened 2015-10-22
142 headless configuration

olaurino:
[release-note]
N/A
[/release-note]

[story] #142 [/story]

# Description
This PR modifies code, pom, and Travis CI configuration to run GUI tests with `UISpec4J` and in headless mode.

## POM changes
The old `NoGuiTests` profiles are removed.

New profiles have been added to the parent project to correctly configure `UISpec4j` and run the GUI tests. Note that `UISpec4j` cannot properly run on all JDKs, so the `ci-headless` profile is enabled by default for convenience.

## Code changes
`NameResolver`, which was replaced long ago by the `HarvardNameResolver` class was removed, along with its test. The test was only running without the `NoGuiTests` profile enabled and it was failing because one of the dependencies (`jel`) had been removed long ago.

The `Position` class, that was original a class nested in `NameResolver`, and that is still used by the name resolver code, has been promoted.

The only GUI test remaining, `BuilderMainViewTest`, has been updated, fixed, and moved to the `iris` package, as it will become an integration/functional test and requires access to the `Iris` class. Some test utility classes were moved to `iris-common` and can thus be used by different modules.

# Testing
The Travis CI builds pass and all the tests, including `BuilderMainViewTest`, are run:
https://travis-ci.org/ChandraCXC/iris/builds/86921024

Unfortunately there is an issue with `openjdk7` that requires a gnome package to be installed, otherwise the Java GUI components trigger the error `GLib-GIO-ERROR **: Settings schema 'org.gnome.system.proxy' is not installed`. I tested locally with `java-1.7.0-openjdk-1.7.0.75.x86_64` on `RHEL`: this should be `openjdk` but I am not sure. Tests pass.
See https://travis-ci.org/ChandraCXC/iris/builds/86919965

# Notes
This PR depends on #162 and should be merged after it.
The commits relevant to this PR are 2ca6a40, a003fa9, f21fa7b.


iris-asdc-plugin
```diff
@@ -0,0 +1 @@
+Subproject commit a40ca4958727cf4962d588ec2a2c1bb6b627f2c8
```
eholum: Can you say a bit about what this file is for?

olaurino:
Please note that this PR allows creating GUI Integration tests following the BuilderMainViewTest prototype. The reason is that on line https://github.com/ChandraCXC/iris/commit/f21fa7bf8fbe2d4ff4690ead87cf6b8308b9e56b#diff-b5a7b9cf67825c906646a0ffe766717fR56 the Tests would need the `Iris.class` class, which is defined in the `iris` module, which is not a dependency for the other modules. The solution would be to add an implementation of the `AbstractIrisapplication` in `iris-common` itself.

If we want to make such changes now I can keep working on this story. Alternatively we can created one more story specifically for this.

iris/src/test/java/cfa/vo/gui/BuilderMainViewTest.java
```diff
@@ -50,52 +41,65 @@
...
-        mainWindow = null;
-        configurationWindow = null;
-    }
+//    @After
+//    public void tearDown() {
+//        mainWindow = null;
+//        configurationWindow = null;
+//    }
 
+    @Ignore
```
eholum: If we put in ignores let's include a comment as to why, so we can remember to put it back in when the why is done.

iris-asdc-plugin
olaurino: git-submodules takes care of keeping track of what commits of a submodule are pointed to by a supermodule. This is the commit this update points to:
ChandraCXC/iris-asdc-plugin@a40ca4958727cf4962d588ec2a2c1bb6b627f2c8

The change in the submodule makes sure that the correct version of `iris-common` is picked up so that the submodule can be included in the final executable/bundle.

sed-builder/src/test/java/cfa/vo/sed/gui/SaveSedDialogTest.java
```diff
@@ -60,8 +61,10 @@ public static void setUpClass() throws Exception {
     
     @AfterClass
     public static void tearDownClass() throws Exception {
-	
+
     }
+
+	@Ignore
```
eholum: Worth commenting on.

Did you switch to tabs instead of spaces? Line indent is off.

sed-builder/src/test/java/cfa/vo/sed/science/stacker/SedStackerRedshifterTest.java
```diff
@@ -113,8 +114,9 @@ public void setUp() {
 	controlX1WithExtraSegment = new double[] {0.90909091, 4.54545455, 9.09090909, 13.63636364, 18.181818182, 45.45454545, 90.90909091};
 	controlY1WithExtraSegment = new double[] {0.1, 0.5, 1.0, 1.5, 2.0, 5.0, 10.0};
     }
-    
-    //@Test
+
+	@Ignore
```
eholum: Same thing with comment and indentations in this guy

iris/src/test/java/cfa/vo/gui/BuilderMainViewTest.java
olaurino: This is weird. The test is indeed being run even if it is ignored. In fact, this test should not be ignored.

I'll look into it.

In any case your comment stands for the rest of the ignored tests. @jbudynk could you please provide me a comment regarding why some of the sedstacker tests are to be ignored?

iris/src/test/java/cfa/vo/gui/BuilderMainViewTest.java
```diff
@@ -50,52 +41,65 @@
...
-        builder.init(new App(), new DesktopWs());
+//        SedBuilder builder = new SedBuilder();
+//        builder.init(new App(), new DesktopWs());
 
         SedlibSedManager manager = (SedlibSedManager) SedBuilder.getWorkspace().getSedManager();
 
-        SedBuilderMainView mainView = new SedBuilderMainView(manager, SedBuilder.getWorkspace().getRootFrame());
+//        SedBuilderMainView mainView = new SedBuilderMainView(manager, SedBuilder.getWorkspace().getRootFrame());
+
+        mainWindow.getMenuBar()
```
olaurino: We should use the WindowInterceptor mechanism here. The way the test is written now is not reliable as it relies on a race condition.

sed-builder/src/test/java/cfa/vo/sed/gui/SaveSedDialogTest.java
olaurino: Interesting, I'll look into it. I did change IDE lately, maybe it's misconfigured.

iris/src/test/java/cfa/vo/gui/BuilderMainViewTest.java
olaurino: I think I know what is going on. The test class is inheriting from `UISpecTestCase` which extends `TestCase`, so the annotations are effectively ignored. It's just a matter of removing the annotations, but we might have issues down the road with this.

jbudynk: This was because sedstacker tests need sherpa connected to the samp hub. At the time, there wasn't an easy way (if any) to get sherpa to connect to the samp hub fast enough before the sedstacker tests ran. 

I just tried running the tests again, with the `//@Test` tags uncommented, and kept getting back a "Sherpa not found" error, along with another SAMP issues:

```
WARNING: Hub ping method failed
java.net.ConnectException: Connection refused
```

jbudynk:
I noticed that there was a test for `NameResolver`. Since `NameResolver` has been replaced with `Position`, do we need a test `PositionTest`?

olaurino:
@jbudynk we do. Should I make it part of this task? I would be happy to, since the class is rather basic :)

jbudynk:
I feel like it'd be a good idea. Since we've agreed when we add new code to the main repository that there's a test for it, perhaps there should be a test :) (even though it's not really "new" code)

iris/src/test/java/cfa/vo/gui/BuilderMainViewTest.java
olaurino: Never mind, the Window Interceptor does not work with JInternalFrames in a Desktop, as I found out during the prototyping phase. I am still concerned that there may be a race condition waiting to be triggered. So far the tests have never failed, but it might be worth researching more into this.

sed-builder/src/test/java/cfa/vo/sed/gui/SaveSedDialogTest.java
olaurino: Some files make a mixed use of spaces and tabs. This is probably due to the different IDEs used. Do we want to enforce a policy on what characters should be used? In that case I would leave this out of the current PR and leave the clean up for another task.

jbudynk: Sounds good to me. I haven't had to deal much with this kind of formatting issue, so either tabs or spaces is fine, as long as we're consistent.

eholum: I vote for spaces, since it looks like the majority of the code already has
spaces.

On Mon, Oct 26, 2015 at 11:18 AM, Jamie Budynkiewicz <
notifications@github.com> wrote:

> In sed-builder/src/test/java/cfa/vo/sed/gui/SaveSedDialogTest.java
> <https://github.com/ChandraCXC/iris/pull/163#discussion_r43006417>:
>
> >      }
> > +
> > +	@Ignore
>
> Sounds good to me. I haven't had to deal much with this kind of formatting
> issue, so either tabs or spaces is fine, as long as we're consistent.
>
> —
> <https://github.com/ChandraCXC/iris/pull/163/files#r43006417>.
>



-- 
Erik Holum

olaurino: I vote for spaces too. Most IDEs should have a setting where you can see characters for spaces and tabs. I'll admit I had it disabled for a while on non-python projects, but it's a good idea to turn it on.

jbudynk: Sounds good. I'll go with the majority rule. Most of the code I've seen
uses 4 spaces per indent, so I vote for 4 spaces, if we need to specify
that here.

We should also make sure that we space correctly when a statement continues
onto the next line. I usually just tab and let the IDE figure it out for me.

On Mon, Oct 26, 2015 at 11:29 AM, Omar Laurino <notifications@github.com>
wrote:

> In sed-builder/src/test/java/cfa/vo/sed/gui/SaveSedDialogTest.java
> <https://github.com/ChandraCXC/iris/pull/163#discussion_r43008052>:
>
> >      }
> > +
> > +	@Ignore
>
> I vote for spaces too. Most IDEs should have a setting where you can see
> characters for spaces and tabs. I'll admit I had it disabled for a while on
> non-python projects, but it's a good idea to turn it on.
>
> —
> <https://github.com/ChandraCXC/iris/pull/163/files#r43008052>.
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

olaurino:
I am closing this PR so we don't lose the comments to the original commits, but I will open a new one after rebasing the underlying branch.

PR #166, opened 2015-10-28
141 setup coverage analysis jacoco

### LINKED ISSUE #141: Setup coverage analysis in build
e.g. cobertura and putting it into the maven website.

We would also like to be able to specify which tests we would like to run, either unit or integration, and break down coverage reports by test type.

olaurino:
[release-note]
N/A
[/release-note]

Resolve #141.

# Description
This PR introduces separate configurations for the unit (`test` phase) and integration (`verify` phase) tests. Coverage analysis plugins are also configured so that coverage information can be created and used for reports. In particular, coverage information is sent to the coveralls service after a successful Travis CI builds.
https://coveralls.io/builds/3975496

The aggregation of the coverage information is performed by coveralls in the `jacoco` profile, and by Sonar in the `sonar` profile, with some configuration tweaks in the `pom` itself.

I also updated the README with information on how to run the different test suites and the coverage analysis plugins.

# Testing
The following Travis Build:
https://travis-ci.org/ChandraCXC/iris/jobs/87897375

resulted in the following Coveralls report:
https://travis-ci.org/ChandraCXC/iris/jobs/87897375

Locally I set up a SonarQube instance and I tested that one can run the whole test suite with the sonar plugin and visualize information in the Sonar web app. The coverage information, in this case, is more accurate and accounts for hits in integration tests from more 

# Caveats
This PR does not refactor tests to be proper integration and unit tests. Historically, most of the Iris tests have been integration tests, and automated GUI testing was not working. Also, some tests are currently ignored as they need to be run locally with the python services running. Similarly, the smoke test only runs with the whole infrastructure up, and again this is not captured by the coverage analysis.
More work will be required to refactor such tests.

JaCoCo and Cobertura do not properly aggregate coverage information and miss the hits coming from integration tests in external jars, even under the same parent module. Sonar properly aggregates the coverage information, but it requires an infrastructure that goes beyond the scope of this PR. Also, the coverage information seems to oscillate with the builds, but the reason is unclear.

For the reasons above, the code coverage identified by coveralls is much lower than the actual code coverage. This is at least twice as much the one advertised by coveralls just by accounting for integration tests outside of `iris-common`. Plus, more coverage is provided by the SedStacker tests that are currently ignored and by the smoke test that is not run as part of the test suite.

The coveralls coverage is actually an estimate of the unit tests coverage.

The `pom` can be configured to allow Sonar to correctly display unit and integration tests coverage disjointly, but we need to evaluate Sonar before spending more work into its configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,4 +27,8 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
 
-script: mvn -U test
+# This runs without instrumentation
+script: mvn -U verify
+
+# This runs with instrumentation
+after_success: mvn -Pjacoco clean verify coveralls:report
--- a/README.md
+++ b/README.md
@@ -13,7 +13,6 @@ $ pip install sampy
 $ pip install astlib
 $ cd sherpa-samp; python setup.py develop; cd ..
 $ cd sedstacker; python setup.py develop; cd ..
-$ mvn test
 ````
 
 You should also make sure that `sherpa-samp` is working.
@@ -30,3 +29,34 @@ $ chmod u+x Iris
 $ ./Iris smoketest
 ````
 
+## How to run the unit and integration tests
+
+### Without coverage analysis
+
+````
+$ mvn clean test # Unit tests
+$ mvn clean test-compile failsafe:integration-test # Integration tests only
+$ mvn clean verify # All tests
+````
+
+### With JaCoCo coverage analysis
+
+````
+$ mvn -Pjacoco test # Unit tests
+$ mvn -Pjacoco verify # All tests
+$ mvn -Pjacoco jacoco:report # generate report
+````
+
+Note that individual reports will be created in each individual submodule.
+
+### With Sonar
+
+A [[http://www.sonarqube.org/ | SonarQube]] instance must be running.
+The configuration for the SonarQube instance must be placed into
+the maven local `settings.xml` file for connecting with the database
+backing the SonarQube instance.
+
+````
+$ mvn -Psonar install # All tests
+$ mvn sonar:sonar
+````
--- a/iris/src/test/java/cfa/vo/gui/BuilderMainViewIT.java
+++ b/iris/src/test/java/cfa/vo/gui/BuilderMainViewIT.java
@@ -45,9 +45,6 @@ public class BuilderMainViewIT extends UISpecTestCase {
 
     public void setUp() throws Exception {
         super.setUp();
-        if (!TestApp.hasComponent(SedBuilder.class)) {
-            TestApp.addComponent(new SedBuilder());
-        }
         setAdapter(new IrisUISpecAdapter(Iris.class, new String[0]));
         mainWindow = getMainWindow();
         desktop = mainWindow.getDesktop();
--- a/iris/src/test/java/cfa/vo/gui/BuilderMainViewIT.java
+++ b/iris/src/test/java/cfa/vo/gui/BuilderMainViewIT.java
@@ -20,6 +20,7 @@
  */
 package cfa.vo.gui;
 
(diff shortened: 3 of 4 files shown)


eholum:
I think they look like good options. I assume the resultant information basically looks the same? Ultimately I think we should just go with whatever is easiest to set up. 

Our goals here are to 

A) Be able to separate unit and integration tests so that we don't have to run all of them all the time.
B) Produce a simple metric to give us an idea of how much of our code is tested.

It looks like A is easily solved by mvn run configurations and good test naming?  

For B, I would say that including integration test from other packages is a secondary concern if it's too much of a headache. Ultimately, I'm not even sure if we care. iris-common is our stand alone, core service, and should be tested independently from the components. Whatever measure of coverage comes from running an integration test on a removeable component shouldn't be counted in the coverage analysis of that package anyway. What happens if we remove it a la specview? So I say we just take the lower values as our starting point, and go from there. 

If we're really wanting a metric to validate that different components work well together, then call pair coverage or something like that is probably the more valuable metric anyway.

The take away here is I think we just merge it and not worry about the caveats.

jbudynk:
I think this is fine to merge. We can run the unit and integration tests
separately, and have the coverage reports setup for at least the unit
tests. I do have some questions:

"JaCoCo and Cobertura do not properly aggregate coverage information and
miss the hits coming from integration tests in external jars, even under
the same parent module."

When building Iris, are we getting the *unit test *coverage for the other
Iris modules, not just iris-common?
Does running mvn -Pjacoco verify not give the proper code coverage for the
whole Iris package?

I think we would want to get coverage from the integration tests for all
the Iris modules together. It sounds like Sonar does the job correctly, and
that we could use that to get our full coverage analysis including
integration tests with another PR. If this is the case, I'd say this PR os
OK to merge.


On Wed, Nov 4, 2015 at 4:41 PM, Erik Holum <notifications@github.com> wrote:

> I think they look like good options. I assume the resultant information
> basically looks the same? Ultimately I think we should just go with
> whatever is easiest to set up.
>
> Our goals here are to
>
> A) Be able to separate unit and integration tests so that we don't have to
> run all of them all the time.
> B) Produce a simple metric to give us an idea of how much of our code is
> tested.
>
> It looks like A is easily solved by mvn run configurations and good test
> naming?
>
> For B, I would say that including integration test from other packages is
> a secondary concern if it's too much of a headache. Ultimately, I'm not
> even sure if we care. iris-common is our stand alone, core service, and
> should be tested independently from the components. Whatever measure of
> coverage comes from running an integration test on a removeable component
> shouldn't be counted in the coverage analysis of that package anyway. What
> happens if we remove it a la specview? So I say we just take the lower
> values as our starting point, and go from there.
>
> If we're really wanting a metric to validate that different components
> work well together, then call pair coverage or something like that is
> probably the more valuable metric anyway.
>
> The take away here is I think we just merge it and not worry about the
> caveats.
>
> —
> <https://github.com/ChandraCXC/iris/pull/166#issuecomment-153875068>.
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

olaurino:
> When building Iris, are we getting the *unit test *coverage for the other Iris modules, not just iris-common?

Yes, we are.

> Does running mvn -Pjacoco verify not give the proper code coverage for the whole Iris package?

It depends on what you mean by "proper". When you run a test from a submodule (the "client") that exercises a different submodule (the "server"), the coverage for the "client" is correctly recorder, while the one for the "server" is not.

Note that there is an open issue regarding this on the jacoco project:
https://github.com/jacoco/jacoco/pull/97

The paradoxical part of it is that this issue is only due to the fact that we have separate submodules. As a matter of fact I was thinking about removing the Berlin Wall among them. That's something we may want to evaluate later on, and I can already think of several pros and cons.

Issue #154, opened 2015-10-21
Evaluate specview units code for our use


eholum:

olaurino:
PR #168 has a lot of working useful for this task. As a matter of fact it might indeed be seen as showing that the specview code could indeed be included, modulo extensive refactoring, into Iris.

However, there are now two places in the code base that have to do with units, and they should be unified: one is package cfa.vo.iris.sed.quantities and one is package cfa.vo.units. Both are in iris-common.

eholum:
We're definitely using it - and there's more work here that needs to be done as @olaurino mentions above. We have units code in two places and they need to be merged. Will track that in new tasks.

For now the units code from specview is already in iris. So calling this guy done.

PR #168, opened 2015-10-28
158 159 160 remove dependency on specview

### LINKED ISSUE #158: Remove the Hack in ExtSED

### LINKED ISSUE #159: Remove Dependency on SPECView and Build Package
Should mostly be done once the replace and stub tasks (156, 157, 160), especially 154 and 159.

### LINKED ISSUE #160: Replace units dependency in iris with a dummy version
Or possibly pulling out the code and copying it into iris or something. Basically need to get the Iris package building without the dependency on specview, so this task is to handle the units portion of that issue.

olaurino:
[release-note]
A new units management and conversion package was built based on the old Specview code.
Physical constants in the units package have been updated to have better precision and to reflect newer measurements of such constants.

In particular:
* H (Planck's constant) now has a value of 6.62606957E-27 erg s.
* C (speed of light) now has a value of 2.99792458E+18 Angstrom/s.
* The MeV to erg conversion factor now has a value of 1.602176565E-6.

[/release-note]

Resolve #158, resolve #159, resolve #160.

# Description
This PR:
* removes all references and dependencies to iris-specview and the Specview jar (#159)
* imports a modified version of the Specview Units package (**check the Copyright is properly noted**)
* replaces the Units calls with a replacement based on the classes coming from Specview (#160)
* The units API is decoupled from the implementation 
* replaces the fit attachment calls with a standalone interface, properly decoupling what was tightly coupled in the current version
* as a result, some hacks have been removed (#158)
* cleans up tab characters in affected classes and removes unused methods
* Introduces a UTYPE class with the one UTYPE that is actually used in the affected code.
* Updates the physical constants to new accepted values, also improving the numerical precision of the constants themselves.
* Replaces the code that took care of parsing multiplication factors in front of unit strings (e.g. '10**3 Jy') to accept any number in floating point notation and different kinds of separators (e.g. '1e3Jy', '1e3*Jy', and '1e3 Jy' are now accepted). Note that the change is not backward compatible, i.e. the old '10**3' notation is not accepted anymore, but this should not be a problem since this code was not really used in Iris, and this change will support future implementation of requirements that we already have.
* Introduces unit tests for the conversion and unit parsing code.

The default implementation is served by a `Default` class with a static member initialized with the default implementation. The actual implementation is also served by the `IWorkspace` interface, and the implementations of this interface are responsible to either instantiate their own units manager or to pick the default one from `Default`.

Unfortunately the only tests that I would expect to be broken, i.e. the ones relevant to SedStacker, are currently ignored, and we don't have tests of the Units framework that we picked off the shelf. Thus, the tests actually pass. The smoke test is broken by a dummy implementation (commit fa31999) while it passes with the specview-based implementation (commit afc01f8).

The smoke test only performs few assertions that rely on units conversions. More unit tests are required.

This PR has implications on PR #154 (Evaluate units code) as it shows the few methods that we actually use from that framework.

I expect more methods to be needed by the Viewer or by a future manager component.

I heavily refactored and simplified the original code. I stopped cleaning up and refactoring the code at some point, mainly for lack of time. There was a lot of unnecessary complexity that could be removed, and some more is still there. We could make more changes as part of #154 and related future tasks.

As for the Fit attachments and the "horrible hack" that this PR removes, one potential refactoring we must evaluate is to provide components vetoing power on the modification of an SED. We had a number of issues in v2.1 that had to do with the fact that specview didn't like (and didn't do much about) users modifying SEDs. While the previous implementation was not perfect, it did indicate that the current Iris framework is lacking the ability of components to veto changes in SED, and to react to such changes. This is a design task left for the future.


olaurino:
@eholum @jbudynk I would like your feedback on this PR before (or together with) a proper code review. In particular:
* do you agree I can put specview back as a dependency for the Units package?
* I don't particularly like that all components have to know about the DummyUnitsFactory implementation and maybe we should start having a common Config object or something like that. Then the individual clients could just take the default implementation from iris-common unless they really want to use a specific implementation (which they don't). It could also go into the Workspace or in the SedManager.

iris-common/src/main/java/cfa/vo/iris/units/XUnits.java
```diff
@@ -0,0 +1,7 @@
+package cfa.vo.iris.units;
+
+/**
+ * Created by olaurino on 10/28/15.
```
eholum: Not important at all, but do we want these comments in here? Ultimately I don't think they add a tremendous amount of value. Class comments should be about the class, and its history is available in git anyway.

iris-common/src/main/java/cfa/vo/iris/utils/UTYPEs.java
```diff
@@ -0,0 +1,8 @@
+package cfa.vo.iris.utils;
+
+/**
+ * Created by olaurino on 10/28/15.
+ */
+public class UTYPEs {
```
eholum: Drops the s

eholum:
It looks good to me. I like have the dummy implementation available to know what we're going to have to be dealing with.

Rather than static singleton instance, I agree with the above and think it would be better to have whatever units functionality we need contained in a class variable of the core iris application (putting it in the workspace sounds good since we're hopefully going to be making changes to the SED manager). My impression is that units conversion is a crucial part of Iris's public API  - and it seems like most callers/components will want to have access to it, and if 3rd party users have some special units package they should be able to pass that in in place of whatever we have as the default. The interface looks simple enough that if somebody did want to provide their own implementation (for some odd reason) that Iris could offer that.

Do we have an idea of how hard it would be to rip SPECView unit conversion classes and just put them in a temporary location in iris-common? Otherwise if it's just the smoke tests/other tests that need it I'm fine leaving the dependency in until we complete a "implement units conversion code" task done. We just need to be sure to keep a task in place to remove it when we have our own package.

And let's talk more about the need for "veto" powers offline. That does make for some interesting complications with the MVC or IOC architecture that's currently in place. We could probably just come up with some kind of state variable in the application that determines that and can be set by individual components when needed - rather than having to get approval from a component when firing off a change notification.

iris-common/src/main/java/cfa/vo/iris/units/XUnits.java
olaurino: That's just the IDE, I guess I stopped paying attention to it long ago, but you are right, we don't want these comments.

jbudynk:
I feel like we wouldn't want to bring back the specview dependency if it's
easy enough to make our own units package or set take the units package
from specview as a dependency. But if either is difficult to do, I agree
with Erik on bringing back the specview dependency for the smoke tests in
the meantime, but making sure we have a task to create a units package to
Iris.

I agree that we should have an interface to the DummyUnits classes; it
makes it easier for 3rd party developers to add their own units
implementation (though they should be very careful with it!).

On Wed, Oct 28, 2015 at 9:18 PM, Erik Holum <notifications@github.com>
wrote:

> It looks good to me. I like have the dummy implementation available to
> know what we're going to have to be dealing with.
>
> Rather than static singleton instance, I agree with the above and think it
> would be better to have whatever units functionality we need contained in a
> class variable of the core iris application (putting it in the workspace
> sounds good since we're hopefully going to be making changes to the SED
> manager). My impression is that units conversion is a crucial part of
> Iris's public API - and it seems like most callers/components will want to
> have access to it, and if 3rd party users have some special units package
> they should be able to pass that in in place of whatever we have as the
> default. The interface looks simple enough that if somebody did want to
> provide their own implementation (for some odd reason) that Iris could
> offer that.
>
> Do we have an idea of how hard it would be to rip SPECView unit conversion
> classes and just put them in a temporary location in iris-common? Otherwise
> if it's just the smoke tests/other tests that need it I'm fine leaving the
> dependency in until we complete a "implement units conversion code" task
> done. We just need to be sure to keep a task in place to remove it when we
> have our own package.
>
> And let's talk more about the need for "veto" powers offline. That does
> make for some interesting complications with the MVC or IOC architecture
> that's currently in place. We could probably just come up with some kind of
> state variable in the application that determines that and can be set by
> individual components when needed - rather than having to get approval from
> a component when firing off a change notification.
>
> —
> <https://github.com/ChandraCXC/iris/pull/168#issuecomment-152046244>.
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

olaurino:
> I like have the dummy implementation available to know what we're going to have to be dealing with.

I guess you mean if we don't actually replace the units code. Otherwise we can remove the dummy implementation, right?

> (putting it in the workspace sounds good since we're hopefully going to be making changes to the SED manager).

Yes, plus it does not really seem to belong to the SedManager. I wonder if it's something that belongs to Sed themselves, though. I can see Seds having a single method convert(String xUnit, String yUnit). This is probably something we can add later on when we review the Sed object model. The implementation of such method would still require to access a more generic implementation, which is useful anyway. I'll go with the Workspace.

> Do we have an idea of how hard it would be to rip SPECView unit conversion classes and just put them in a temporary location in iris-common?

We have a task for that. I can include that in my work too. If it's easy enough I can just do it, but the bulk of the actual implementation, in that case, would be writing unit tests. Or we could keep the original task as an evaluation of how the code should be refactored after its inclusion.

> if it's just the smoke tests/other tests that need it

For now yes, but it's also Jamie's SedStacker components/tests that rely on units conversions, as soon as we unIgnore them (which I hope is soon enough).

> And let's talk more about the need for "veto" powers offline.

Indeed, that's what I am calling for now that the "horrible hack" is out of the way. The use case is that a user may e.g. delete a segment from a SED in the Builder. The Builder does not care about the consequences of the action. However, other components, e.g. the fitting tool, may notify the users that if they change the SED something is going to happen to their Fit, e.g. in the simplest case it is going to become outdated. In that case the user might decide to cancel the action. This veto (the word might be too strong) must come before the action is actually performed, and more events may be triggered at the same time. Again, this is material for a new task, probably way down the road.

iris-common/src/main/java/cfa/vo/iris/utils/UTYPEs.java
olaurino: :+1:

olaurino:
I updated the implementation by taking into account your suggestions (I removed some `author` comments but many are still in there).

I updated the PR description.

In particular, note the way the units manager is served by the IWorkspace interface.

Porting the Specview units package was not too hard, and the implementation fixes the smoke test broken by earlier commits. We should add some serious unit testing for these classes, though, and most of their code is not even used right now. Actually, I am thinking of commenting out all the unused code.

olaurino:
This PR is ready to be reviewed. Note that despite the original tasks said, it actually implements a full blown replacement for the old code, by embedding the original specview code and then modifying it (in some cases rather significantly).

jbudynk:
Minor note:

Does it matter if the <p> and <li> tags in the comments are closed or not?
E.g., iris-common/src/main/java/cfa/vo/iris/units/spv/UnitStringParser.java
 lines
34-41 close the <p> tags </p>, where as
iris-common/src/main/java/cfa/vo/iris/units/XUnit.java lines 29-53 do not
close the <li>, <ul> or <p> tags.

On Tue, Nov 10, 2015 at 9:52 AM, Omar Laurino <notifications@github.com>
wrote:

> This PR is ready to be reviewed. Note that despite the original tasks
> said, it actually implements a full blown replacement for the old code, by
> embedding the original specview code and then modifying it (in some cases
> rather significantly).
>
> —
> <https://github.com/ChandraCXC/iris/pull/168#issuecomment-155440584>.
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

iris-common/src/main/java/cfa/vo/iris/desktop/IrisWorkspace.java
```diff
@@ -39,6 +43,8 @@
 
     private ISedManager sedManager;
 
+    private static final JFileChooser fileChooser = new JFileChooser();
```
eholum: What's this used for?

iris-common/src/main/java/cfa/vo/iris/sed/ExtSed.java
```diff
@@ -59,6 +59,7 @@
     private Map<String, Object> attachments = new TreeMap();
     private String id;
     private boolean managed = true;
+    private static UnitsManager uf = Default.UNITS_MANAGER;
```
eholum: Does this need to be a private static field? 

If the units manager is a class variable in the workspace, then it should be the workspace that passes the instance of the units manager to each SED that needs it when it needs it. Otherwise when we override the default behavior in the workspace we might end up with different behavior in the SED objects.

iris-common/src/main/java/cfa/vo/iris/units/DefaultUnitsManager.java
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+package cfa.vo.iris.units;
+
+import cfa.vo.iris.units.spv.UnitsFactory;
+import cfa.vo.iris.units.spv.VelocityUnits;
+
+public class DefaultUnitsManager implements UnitsManager {
+
+    public DefaultUnitsManager() {
```
eholum: Is this supposed to be a singleton? It's not locked down but it looks like everyone is just accessing a static instance in the Default class.

If we want it to be a singleton let's stick with a standard pattern. I'm a fan of https://en.wikipedia.org/wiki/Singleton_pattern#The_solution_of_Bill_Pugh

If we want the Default class it should just reference that instance. Might also be worth putting methods in the default class as opposed to statics, to make things more easily "stubbable".

sed-builder/src/main/java/cfa/vo/sed/science/stacker/SedStack.java
```diff
@@ -1,12 +1,12 @@
 /**
```
eholum: Yay! There's going to be some duplicate reformatting of spaces in https://github.com/ChandraCXC/iris/pull/179

olaurino:
@jbudynk:
> Does it matter if the <p> and <li> tags in the comments are closed or not?

I don't think it does, but this is one more reason to include the javadocs report in the site. I may have done other mistakes as well.

iris-common/src/main/java/cfa/vo/iris/desktop/IrisWorkspace.java
olaurino: At some point a user requested that the file chooser would have the same state across different components. One way of doing it (crude but quick and easy to use on the clients) was to store the filechooser instance itself in the workspace.

iris-common/src/main/java/cfa/vo/iris/sed/ExtSed.java
olaurino: it does not and it was not supposed to be static. Thanks for catching it.

olaurino: Sorry, it needed to be static because it is used by static methods offered by ExtSed.

iris-common/src/main/java/cfa/vo/iris/units/DefaultUnitsManager.java
olaurino: No, it was not supposed to be a singleton, although it might be.

I created the Default class to hold instances that should be shared by components unless they want to do otherwise, without requiring the components to know what implementation they are using. On the other hand The ExtSed class does not know anything about the workspace (as a matter of fact it should be a class in sedlib), and thus needs to instantiate the default units manager itself. Making the class a singleton would not help here. Since the ExtSed class and the components both need a units manager, it may just be that the units manager should be defined either in a third component or in sedlib itself.

In any case I will change the Default implementation to use a method rather than a static field. Note that Default does need to be a Singleton in that case.

olaurino:
@eholum please let me know if commit 27de071 addressed your comments.

eholum:
Looks good.

On Thu, Nov 12, 2015 at 3:06 PM, Omar Laurino <notifications@github.com>
wrote:

> @eholum <https://github.com/eholum> please let me know if commit 27de071
> <https://github.com/ChandraCXC/iris/commit/27de0710cd521d48982cbcbfa5545287bedc2d0a>
> addressed your comments.
>
> —
> <https://github.com/ChandraCXC/iris/pull/168#issuecomment-156219748>.
>



-- 
Erik Holum

PR #178, opened 2015-11-06
Logging for Unit Tests Cleanup

eholum:
Building off Omar's change.

Logs for units tests are dumped into surefire output. 
Removed most of the prints to std out - those that are left are only in the CLI and visible for Consumers' sake.
Formatted logfiles for readability, no longer split across lines, e.g.:

2015-11-06 11:48:32 cfa.vo.iris.ComponentLoader initComponents [INFO] Loading class: cfa.vo.iris.ComponentLoaderTest$BrokenTestIrisComponent
2015-11-06 11:48:32 cfa.vo.iris.ComponentLoader initComponents [SEVERE] Could not construct component cfa.vo.iris.ComponentLoaderTest$BrokenTestIrisComponent
java.lang.RuntimeException: I don't work
	at cfa.vo.iris.ComponentLoaderTest$BrokenTestIrisComponent.<init>(ComponentLoaderTest.java:73)

Ultimately I think we should setup an option for specifying log files for the Iris application VIA the CLI or the preferences config file.

All tests pass, verified that log files have appropriate format, travis builds are much easier to read. Failed tests/processes still dump their exceptions into the log.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/AbstractIrisApplication.java
+++ b/iris-common/src/main/java/cfa/vo/iris/AbstractIrisApplication.java
@@ -43,57 +43,51 @@ import org.astrogrid.samp.client.SampException;
 import org.jdesktop.application.Application;
 
 /**
+ * Base Iris application. Handles startup, shutdown, etc.
  *
- * @author olaurino
  */
 public abstract class AbstractIrisApplication extends Application implements IrisApplication {
-
-    private static SedSAMPController sampController;
+    
+    private static final Logger logger = Logger.getLogger(AbstractIrisApplication.class.getName());
+    
     private static boolean isTest = false;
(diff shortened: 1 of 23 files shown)


olaurino:
Strangely enough one of the travis build failed. @eholum could you please check why?

eholum:
Yeah, basically boils down to a race condition between the test runner, the iris desktop, and uispec4j. As of now there's no guarantee that the desktop window has reset to required start conditions before junit executes each test - leaving some probability of each desktop component not being available.

I don't think there's much to do other than put in a wait... in the test setup. Which makes me kind of sad. Otherwise just restarting the build when this breaks... which also makes me even more sad.

eholum:
Okay, line coverage is increased in here:

https://github.com/ChandraCXC/iris/pull/179

Issue #144, opened 2015-10-21
Add images to documentation


eholum:
Add existing images to doc

PR #182, opened 2015-11-12
Docs migration

jbudynk:
Address #144, #150

# Description

This PR:

   - migrates the current Iris v2.1 user documentation from the home-grown CXC site building system to Maven site.
   - adds very simple sites for each project module so that developers can view the dependencies and other build information. E.g., `sed-builder/src/site/`.
   - adds a new repository, [iris-docs](https://github.com/jbudynk/iris-docs), as a submodule to the Iris project. This contains all the Iris user documentation and images (#144).
   - the site can now be pushed to versioned sites on the CXC server using `mvn site:deploy` (#150). This requires the CXC server to be added to the developer's `settings.xml` file. The site directory name is ${project.version}.

The site build still needs the following additions before completion:

   - create profiles to push the site to the CXC test and live sites.
      - write instructions to pushing the site to the preview and live servers.
   - fix repository blacklist warnings. When running `mvn site:site`, a number of repositories that don't exist are shown to be blacklisted. I don't know what's causing these warnings, but these should be understood before we release the site to the live server.
   - remove redundant dependencies. The project module reports show a few redundant dependencies. 
   - fix typo in project modules' `site.xml`. "Sybling Modules" should be "Sibling Modules."

Another issue to tackle is that the `iris-docs` repo is under my account, not ChandraCXC. We should address this before merging the PR.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "sedlib"]
 	path = sedlib
 	url = https://github.com/ChandraCXC/sedlib
+[submodule "iris-docs"]
+	path = iris-docs
+	url = https://github.com/jbudynk/iris-docs.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -15,4 +15,4 @@
 	url = https://github.com/ChandraCXC/sedlib
 [submodule "iris-docs"]
 	path = iris-docs
-	url = https://github.com/jbudynk/iris-docs.git
+	url = https://github.com/ChandraCXC/iris-docs.git
--- a/iris-common/pom.xml
+++ b/iris-common/pom.xml
@@ -29,6 +29,11 @@
         <version>3.0-SNAPSHOT</version>
     </parent>
 
+    <!-- To help with site creation directory, specify the parent directory -->
+    <properties>
+        <main.basedir>${project.parent.basedir}</main.basedir>
+    </properties>
+
     <build>
         <plugins>
             <plugin>
@@ -42,6 +47,26 @@
                         </goals>
                     </execution>
                 </executions>
+            </plugin>
+            
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-site-plugin</artifactId>
+                <version>3.4</version>
+                <configuration>
+                    <siteDirectory>./src/site/</siteDirectory>
+                    <outputDirectory>${project.parent.basedir}/target/site/iris-common/</outputDirectory>
+                </configuration>
+<!-- 
+                <dependencies>
+                    <dependency>
+                        <!~~ add support for ssh/scp ~~>
+                        <groupId>org.apache.maven.wagon</groupId>
+                        <artifactId>wagon-ssh</artifactId>
+                        <version>1.0</version>
+                    </dependency>
+                </dependencies>
+-->
             </plugin>
         </plugins>
     </build>
@@ -104,5 +129,4 @@
             <version>0.9.7</version>
         </dependency>
     </dependencies>
-
 </project>
--- a/iris-common/pom.xml
+++ b/iris-common/pom.xml
@@ -29,11 +29,6 @@
         <version>3.0-SNAPSHOT</version>
     </parent>
 
-    <!-- To help with site creation directory, specify the parent directory -->
-    <properties>
-        <main.basedir>${project.parent.basedir}</main.basedir>
-    </properties>
-
     <build>
         <plugins>
             <plugin>
@@ -47,16 +42,6 @@
                         </goals>
                     </execution>
                 </executions>
-            </plugin>
-            
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-site-plugin</artifactId>
-                <version>3.4</version>
-                <configuration>
-                    <siteDirectory>./src/site/</siteDirectory>
-                    <outputDirectory>${project.parent.basedir}/target/site/iris-common/</outputDirectory>
-                </configuration>
             </plugin>
         </plugins>
     </build>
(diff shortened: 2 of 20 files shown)


jbudynk:
There are a few TODO's in the `iris-docs` submodule as well:

   - an image on the download/index.md page isn't showing up (right before the "Smoke Test" section). There should be an image of the Iris desktop there.
   - move the <a name="top"></a> anchors above the header element instead of in it; otherwise, the HTML code is copied directly to the site title name.
   - replace hard-coded "v2.1" to "${project.version}" in the Markdown documentation.

olaurino:
I pushed some commits addressing the following comments:

* README:
  * the statement that the documentation is based on Iris 2.1 is probably redundant and won't make sense pretty soon.
  * the "open" command is platform-dependent, I believe, i.e. it would work on Macs only.
  * the "site:run" command is rather useful but it is not included in the README: it runs an embedded web server that updates the files every time they are changed: you make a change in the .md file, you save it, reload the page and jetty will render the edited page.
  * the "site:site" command does not (by design) install the submodules websites. This can be achieved with "site:stage". However, as far as I can tell there is no way to run jetty ("site:run") on the staging folder. I haven't tried too hard, though. This should be noted in the README.

* pom configurations can be greatly simplified:
  * the site plugin *declaration* in the parent pom is inherited by the children, so this does not need to be repeated in all the children's poms, where the default configuration is enough.
  * the parent pom's site plugin *configuration*, on the other hand, should not be inherited, and this can be done in the parent's pom itself, without requiring the children to override it.
  * I am not too fond of including our workstations IP addresses in a public pom. Plus, it looks like maven echoes the password in the consoles when scp:// is used. I know Jamie had issues with using the file:// protocol, but I think I found the reason and a solution to that (see below). SCP should only be used with a private key or maybe with a secured settings.xml (proper permissions, obscured password). Since the local protocol works, let's stick to that for the time being. I would also create profiles for staging and deploying the site in the production area.

Something else bothers me in the site configurations, i.e. that the site.xml is duplicated in all of the submodules. This should (and can, as far as I know) avoided because it's error-prone. However, the documented solutions do not work, or I can't understand how to make them work.

Also, we should add the javadocs, at least for some of the submodules like iris-common and samp-factory that expose an API used by the other submodules. Without the javadocs the submodule website do not seem to make too much sense if not for properly accounting for the dependencies. By the way, I would remove the dependencies report from the parent module altogether, but let's keep it there for the moment.

I also have a question: I noticed that you added the mycila-license plugin to the submodule. Was this required? As far as I can tell that plugin only introduces a command that can install missing copyright headers. Or am I missing something?

olaurino:
I also created a new iris-docs repo (I'll make sure the Iris team is assigned to it in a moment), and updated the git submodule configuration to point at that repo.

jbudynk:
I will remove the emails and the machine address (under distributionManagement) from the POM. Omar and I had talked about this earlier too.

jbudynk:
I am unable to run `mvn site:stage` now, from either my laptop or workstation. It's complaining about not having a distributionManagement section in the POM:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.4:stage 
(default-cli) on project iris2: Missing distribution management in project Iris
```

I'll work on fixing this.

jbudynk:
Alright, so I've added a few changes to the doc-related files:

   - I put spaces between the email parts, like "jbudynkiewicz @ cfa.harvard.edu
   - I introduced the <distributionManagement> section back into the main POM so that site:stage works. By default, the website built with site:stage is stored in `target/staging`, despite the fact that the <url> element says to store it at the preview CXC server. One can specify the directory to store the staged site at using `-DstageLocation=/path/to/directory`
   - Updates the iris-docs module to include typo fixes and fixes the "#top" anchors on each page. For background: each page had an anchor for going back to the top of a page. The anchor was written in-line with the page header in Markdown like

        # <a name="top"></a> Thread Title

   Before commit 3d02ba4, hovering over the site tab in the web browser showed "`Iris - <a name="top"></a> Thread Title`" instead of "`Iris - Thread Title`." I moved the `<a name="top"></a> to the `<head>` block in `iris-docs/site.xml` so that each page has a top anchor.
   - The Iris desktop image in `download/index.md` that was supposedly missing was just extraneous Markdown syntax that tried to center the image, but it didn't work. I removed the extraneous lines.
   - I removed the unnecessary mycila-license-plugin from `iris-specview/pom.xml`.

jbudynk:
Alright, the PR's ready to be merged. I added the SI header metadata to each `site.xml` (#184), including the project modules'. I can remove these from the project modules if you wish. 

@olaurino @eholum  any final thoughts on the latest additions?

olaurino:
I just wish I had figured out how to inherit site.xml... having to mirror each change in each submodule is not sustainable and error-prone. As long as they are the guaranteed to be the same, maybe we could use the resources plugin to copy the same file over each individual submodule, but that's for another day, and in any case it would be a workaround too. I think we can merge it.

jbudynk:
Agreed. That would make it much less error prone. For now, I've just been
editing one site.xml and copying it to the rest (except
iris-common/src/site/site.xml, which has some commented out code I'd like
to keep for now).

On Fri, Nov 13, 2015 at 2:44 PM, Omar Laurino <notifications@github.com>
wrote:

> I just wish I had figured out how to inherit site.xml... having to mirror
> each change in each submodule is not sustainable and error-prone. As long
> as they are the guaranteed to be the same, maybe we could use the resources
> plugin to copy the same file over each individual submodule, but that's for
> another day, and in any case it would be a workaround too. I think we can
> merge it.
>
> —
> <https://github.com/ChandraCXC/iris/pull/182#issuecomment-156538794>.
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

eholum:
Looks good to my eyes and on my box. We can add a backlogged item to come back and look at inheriting xml from parents with the other documentation items. I say merge it.

PR #197, opened 2015-11-18
Iris 174 run smoketest

### LINKED ISSUE #174: Make smoketests run as apart of the integration tests

olaurino:
Resolve #174.

Run the smoke test as part of the maven's integration-test phase.

Some refactoring was required to make sure that a CLI of an IrisComponent could set the output state of the shell process, and that this status could bubble up in the Iris launcher script. This way an error in the smoke test would fail the maven build. I tested this locally by faking a failure in the smoke test itself.

Some clean up was performed by removing unused code (a test builder that was not used anymore) and removing some comments. This clean up was kept in this PR because it is related to the smoke test or because the affected classes were to be changed as part of the refactoring.

Also, I made sure that both the commits in this PR are green, at least locally.

Note that as part of #188 the smoke test execution might be moved to the post-integration-phase, or even to the verify phase. This is because the smoke test actually sets up its own environment, including launching SAMP, and when the integration tests infrastructure will set up SAMP we don't want to be dependent on the order in which things are run. While I think the smoke test would just use the running SAMP hub instead of starting one itself, we would fail to make sure that the smoke test runs by itself.

I could move the smoke test to the verify phase as part of this PR's review if needed.

This build shows that the smoke test is being run:
https://travis-ci.org/ChandraCXC/iris/jobs/91848840

Something else to keep in mind for the review: the smoke test's output is included in the maven output. I think this makes sense because of the nature of the smoke test. However, we can disable it by configuring the maven exec plugin to redirect stdout and stderr to a file.


iris/src/main/java/cfa/vo/iris/smoketest/SmokeTest.java
```diff
@@ -76,7 +75,7 @@ public void call(String[] args) {
                 } catch (Exception ex) {
                     Logger.getLogger(SmokeTest.class.getName()).log(Level.SEVERE, null, ex);
                 } finally {
-                    test.exit();
+                    return test.exit();
```
eholum: https://github.com/ChandraCXC/iris/pull/197/files#diff-6d6e7861b637165c4d71bf5e97cd5c1aR249 catches and logs a throwable, it should pass the throwable up to here and set an failing exit status. Or just something so that the sherpasmoketests fail if something breaks.

jbudynk: I agree - it makes sense to have the smoketest start a new SAMP Hub. While
it's less efficient, I think we'd want to run the smoketests under the same
or as close to the same conditions as the user will run them.


And other interesting info:

Smoketests pass when you have a SAMP hub already running. I unknowingly had
a SAMP hub open on my laptop, but noticed before the smoke tests ran. I saw
that a new Sherpa instance connected to the Hub. Most of the messages were
being sent to the new Sherpa instance, but a few responses were sent to the
old Sherpa instance.


On Fri, Nov 20, 2015 at 12:14 PM, Erik Holum <notifications@github.com>
wrote:

> In iris/src/main/java/cfa/vo/iris/smoketest/SmokeTest.java
> <https://github.com/ChandraCXC/iris/pull/197#discussion_r45492917>:
>
> > @@ -76,7 +75,7 @@ public void call(String[] args) {
> >                  } catch (Exception ex) {
> >                      Logger.getLogger(SmokeTest.class.getName()).log(Level.SEVERE, null, ex);
> >                  } finally {
> > -                    test.exit();
> > +                    return test.exit();
>
>
> https://github.com/ChandraCXC/iris/pull/197/files#diff-6d6e7861b637165c4d71bf5e97cd5c1aR249
> catches and logs a throwable, it should pass the throwable up to here and
> set an failing exit status. Or just something so that the sherpasmoketests
> fail if something breaks.
>
> —
> <https://github.com/ChandraCXC/iris/pull/197/files#r45492917>.
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

jbudynk:
I agree - it makes sense to have the smoketest start a new SAMP Hub. While
it's less efficient, I think we'd want to run the smoketests under the same
or as close to the same conditions as the user will run them.

Looks good to me.

And other interesting info (unrelated to my statements above):

Smoketests pass when you have a SAMP hub already running. I unknowingly had
a SAMP hub open on my laptop, but noticed before the smoke tests ran. I saw
that a new Sherpa instance connected to the Hub. Most of the messages were
being sent to the new Sherpa instance, but a few responses were sent to the
old Sherpa instance.

iris/src/main/java/cfa/vo/iris/smoketest/SmokeTest.java
olaurino: Good catch, I'll fix it.

olaurino:
Rebasing: original tip was at  456f7b2

olaurino:
commit ea094ed addresses @eholum's comment https://github.com/ChandraCXC/iris/pull/197#discussion_r45492917. @eholum do you agree? I tested locally that throwing a fake exception the smoke test failure is properly logged and the build is properly failed.

Also, commit 68asf37 makes the smoke test run during the verify phase rather than integration-test.

eholum:
Yep looks good.

olaurino:
Closing this PR as it was replaced by #207

PR #198, opened 2015-11-18
Fix docs migration

jbudynk:
# Description

This PR is in response to PR #182. Not all the necessary changes were added to the docs migration in PR #182, so I've added the following updates to fix this:

   - I put spaces between the email parts, like "jbudynkiewicz @ cfa.harvard.edu". Please let me know if you'd prefer me to just get rid of them, or if we should just provide irisdev as the email, or anything else. Also, I removed Janet's email because I figure she wouldn't want people emailing her about Iris (and we'd want to get the emails first anyhow) - let me know if I should add it back in.
   - I introduced the <distributionManagement> section back into the main POM so that site:stage works. By default, the website built with site:stage is stored in `target/staging`, despite the fact that the <url> element says to store it at the preview CXC server. One can specify the directory to store the staged site at using `-DstageLocation=/path/to/directory`
   - Updates the iris-docs module to include typo fixes and fixes the "#top" anchors on each page. For background: each page had an anchor for going back to the top of a page. The anchor was written in-line with the page header in Markdown like

        # <a name="top"></a> Thread Title

   Before commit 88baf67, hovering over the site tab in the web browser showed "`Iris - <a name="top"></a> Thread Title`" instead of "`Iris - Thread Title`." I moved the `<a name="top"></a> to the `<head>` block in `iris-docs/site.xml` so that each page has a top anchor.
   - The Iris desktop image in `download/index.md` that was supposedly missing was just extraneous Markdown syntax that tried to center the image, but it didn't work. I removed the extraneous lines.
   - I moved / renamed `iris-specview/src/site` to is replacement, `iris-visualizer/src/site`.

If needed, review PR #182 for the rest of the docs migration details. 

Note that I cherry-picked commits from the deleted branch `docs_migration` (was at commit bb9296c) to `fix_docs_migration`.

### DIFF
--- a/iris-common/src/site/site.xml
+++ b/iris-common/src/site/site.xml
@@ -13,7 +13,9 @@
     
         <head>
 	    <!-- meta keywords for SI / HEA -->
-            <meta http-equiv="keywords" content="chandra, axaf, x-ray, astronomy, astrophysics, catalog, multi-wavelength, multiwavelength, vao, VAO, sed, SED, SEDs"/>
+            <meta http-equiv="keywords" content="chandra, axaf, x-ray,
+            astronomy, astrophysics, catalog, multi-wavelength,
+            multiwavelength, vao, VAO, sed, SED, SEDs, spectral energy distribution"/>
             <meta http-equiv="description" content="Iris: VO SED Analysis Tool"/>
             <meta name="creator" content="SAO-HEA"/>
             <meta name="keywords" content="SI,Smithsonian,Smithsonian Institute"/>
--- a/iris-common/src/site/site.xml
+++ b/iris-common/src/site/site.xml
@@ -11,6 +11,16 @@
 
     <body>
     
+        <head>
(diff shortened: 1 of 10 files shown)


eholum:
Yeah looks good to me. Only think I could think of is adding "Spectral Energy Distribution" specifically to the keywords if this site if we want crawlers to pick up on those, too. Otherwise site looks goo on my box, still.

jbudynk:
Good idea - and adding this to the PR helped me find a typo in one of my other files, so thanks!

One more issue popped up though, and I think it stems from the PR #182. The project submodule intro pages don't show the titles correctly. For example, instead of saying "Iris - iris-visualizer", it's just "Iris - ". All other page titles (submodule or user docs) look fine.

jbudynk:
Alright, in my latest change, I updated the email addresses in the main POM so they read like

username_at_address

Let me know if anything else should be changed.

olaurino:
Merge it :)

PR #199, opened 2015-11-18
189 update doc strings

### LINKED ISSUE #189: Document Unit Testing Iris Components
Now that the abstract GUI testing infrastructure is in place.

olaurino:
Resolve #189 

This PR updates the docstrings for the abstract gui unit test classes.

More information for the developer (which may end up in the site when we review the developer documentation) is in the internal UDF. See email to internal mailing list.

I will be happy to include such documentation in the project now, but I will need to research how to do it, which may be a new task.


eholum:
Looks good, ditto on the UDF notes -

olaurino:
There is an issue in this PR and the accompanying UDF document, as the JUnit 4 annotations are ignored since the abstract class extends the uispec abstract test case, which in turn extends JUnit's test case. I'll see if it makes sense to update the code since we are changing it anyway, but I'll certainly update the UDF document.

olaurino:
Rebasing on top of the latest master. The original commit was a39ce3f.

Issue #191, opened 2015-11-13
Implementation of an Abstract Test for Integration Testing


eholum:
Needs to verify Sherpa Samp/SAMP and everything else that needs to be setup is setup

PR #200, opened 2015-11-24
Iris 187 188 start sherpa samp for it

olaurino:
Address #187 and address #188 

This PR instructs maven to start a samp hub and sherpa-samp as external processes in the pre- and post-integration test maven phases. More work needs to be done as issues came up during tests (see Coming Soon section below).

# Notes
I started working on task #188 and then realized that #187 was pretty similar to it and easy to implement and test in the changes I was already making. Since Jamie hadn't start working on it I prototyped something for #187 as well.

I would still consider this as a prototype, especially because we still don't have the IntegrationTest abstract classes (suites?) in place. However, one of the several successful tests was to see the old BuilderMainViewIT integration test fail because the SAMP window did not show up. This is expected (even though I had forgot about it) because we don't need the SAMPUISpecAdapter anymore: the samp hub will be launched by maven in the pre-integration-test phase and there is no need for a special adapter to intercept the SAMP hub window.

So, with the above commit I also removed the unnecessary adapter, and I modified the BuilderMainViewIT to extend the AbstractGUIComponent class.

# Testing
I did several tests locally on my workstation, but here is the proof of the configuration working on Travis CI:
https://travis-ci.org/ChandraCXC/iris/jobs/91914863

For the record, here is the build that failed because the SAMP hub window did not show up:
https://travis-ci.org/ChandraCXC/iris/jobs/91910559

Another commit piggy backs the one above as I replicated the logging configuration defined for surefire (unit tests) on failsafe (integration tests), so integration tests do not log during the maven execution.

# Coming Soon
The sed-builder does not have any integration tests (yet), so the ant-run plugin execution should be removed from its pom.

Also, killall fails on OSX, so I'll replace it with `pkill -f` which seems to work the same on Linux and OSX.


eholum:
Looks good. Verified that it pulls down and correctly runs on my machine. I think that caveat with killing the samp pid we talked about is fine - and more of an issue for the smoketests covered in https://github.com/ChandraCXC/iris/issues/201

Since this will be run before and after all integration tests - we don't have to worry about stubbing anything out for GUI integration tests, as this will already be there.

Thumbs up for merging.

jbudynk:
I've based my changes for #191 on this PR and it seems to be working as expected.

Looks good to me

olaurino:
Rebasing because of conflicts. Original tip was at 01f2b54.

olaurino:
Also, added a fix for pkill vs killall (tbh I haven't tested on osx just yet). Also, in order to make the logs more clear I updated the ant tasks names.

eholum:
Looks good. Still behaving as expected on my linux machine, though I haven't tested it on mac either.

olaurino:
Closing this PR as it has been replaced by #207.

Issue #172, opened 2015-10-29
Remove @Ignore from SED Stacker Tests


eholum:
Once units are in place and things will actually work.

olaurino:
It looks like some tests were hanging not quite because the window were shown, but because UISpec4j was not initialized. When one initializes UISpec4J it will fail when the windows are shown if they were not supposed to be shown. I think this is the correct behavior, and we don't need to do anything with the windows themselves. Nevertheless, I did try to intercept the error dialog box and dismiss them. In that case the tests fail with the exception thrown by the sedstacker, which says that sherpa could not be found.

The first implementation (let UISpec4J fail when an undesired error message is shown) is much simpler and the error message is very clear. The second option (getting rid of the window) is a little bit cleaner but unnecessarily complicated.

@eholum, @jbudynk, do you think I can proceed with option 1?

jbudynk:
Yes, the test should fail if an expected pop-up window isn't shown.

On Wed, Nov 25, 2015 at 12:08 PM, Omar Laurino <notifications@github.com>
wrote:

> It looks like some tests were hanging not quite because the window were
> shown, but because UISpec4j was not initialized. When one initializes
> UISpec4J it will fail when the windows are shown if they were not supposed
> to be shown. I think this is the correct behavior, and we don't need to do
> anything with the windows themselves. Nevertheless, I did try to intercept
> the error dialog box and dismiss them. In that case the tests fail with the
> exception thrown by the sedstacker, which says that sherpa could not be
> found.
>
> The first implementation (let UISpec4J fail when an undesired error
> message is shown) is much simpler and the error message is very clear. The
> second option (getting rid of the window) is a little bit cleaner but
> unnecessarily complicated.
>
> @eholum <https://github.com/eholum>, @jbudynk <https://github.com/jbudynk>,
> do you think I can proceed with option 1?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ChandraCXC/iris/issues/172#issuecomment-159675152>.
>



-- 
Jamie A. Budynkiewicz
Smithsonian Astrophysical Observatory
Harvard-Smithsonian Center for Astrophysics
100 Acorn Park Dr. R-377 MS-81
Cambridge, MA 02140

PR #202, opened 2015-11-25
Iris 172 unignore stacker tests

olaurino:
Address #172 

Depends on #200 

# Description
This PR enables the sed stacker tests that are currently ignored because they need a samp hub and sherpa-samp running.

This PR now depends on #200, and Travis builds should pass. However, a previous edition of this PR did not depend on #200, so there were no samp hub and sherpa-samp available. In this case having Travis builds fail is a test of its own, since it shows that the tests fail for the correct reason when they cannot find sherpa in the hub. Also, they don't hang:
https://travis-ci.org/ChandraCXC/iris/jobs/93219011

For the record the original tip of this PR was at b677802. The commits relevant to this PR are 7fb49c4, b9a4e95, 1dc1dd7, and ca89bca.

Currently there is one more ignored test (SherpaRedshifterTest). Not sure whether I should unignore it at this stage as this would require making it somehow part of the current sedstacker tests infrastructure. Maybe we should just wait for the new infrastructure to be in place at this end of this sprint... then it should trivial to unignore that test as well.


olaurino:
Rebasing onto #200. The Travis builds should now pass.

Some changes were required when integrating the code from #200, as the abstract test class would not wait for sherpa to show up. This required a (long overdue) addition to SherpaClient as well, and a minor refactoring of the sed stacker integration tests.

olaurino:
It looks like the travis builds are hanging. Not sure why, I'll have to look into it after the holiday.

olaurino:
There were a couple of issues:
  * I forgot to make a timeout variable actually run, so the tests hang in an infinite loop under some conditions
  * The `waitFor` method waited for sherpa-samp to connect to the hub, but not for it to be *pingable*, so that a science RPC could be sent before sherpa-samp had declared its subscriptions.

Both issues should be fixed now.

iris-common/src/main/java/cfa/vo/sherpa/SherpaClient.java
```diff
@@ -160,6 +161,42 @@ public String findSherpa() throws SampException {
         }
     }
 
+    public boolean waitFor(int timeoutMillis, int step) {
+        String sherpaId = null;
+        int elapsed = 0;
```
eholum: This function could end up waiting for much greater than 2x timeoutMillis owing to elapsed being reset inbetween finding/pinging sherpa and to drift - particularly from the callAndWait function on the SAMPController. Would rather see elapsed as something like

start = System.currentTimeMillis();
elapsed  = System.currentTimeMillis() - start;

To give callers a more reliable timeout.

iris-common/src/main/java/cfa/vo/sherpa/SherpaClient.java
```diff
@@ -160,6 +161,42 @@ public String findSherpa() throws SampException {
...
+            }
+            elapsed += step;
+        }
+
+        boolean sherpaResponded = false;
+        elapsed = 0;
+
+        while (!sherpaResponded && elapsed < timeoutMillis) {
+            try {
+                this.sampController.callAndWait(sherpaId, new PingMessage().get(), step/1000);
```
eholum: Does it matter that there will be no timeout if step < 1000?

iris-common/src/main/java/cfa/vo/sherpa/SherpaClient.java
olaurino: I have to say I didn't care too much about the accuracy of the timeout. Semantically, one might say that the timeout argument is applied to two actions rather than one. As for the drift, I was aware of that but I thought I didn't care enough. It's easy to clean it up, so I'll implement your suggestion.

I am wondering, however, if we should just use the Future API. We could use it here where it is simple enough, and then reuse the pattern for more complicated cases, i.e. when fitting (you may want to cancel a long-running fit or the computation of confidence levels). I'll look into it.

iris-common/src/main/java/cfa/vo/sherpa/SherpaClient.java
olaurino: Argh. This was not intended. However, is it possible that one may want to run without a timeout? Truth is we are mixing ints with floats. I'll try to clean up.

iris-common/src/main/java/cfa/vo/sherpa/SherpaClient.java
eholum: I think you're right in that it doesn't matter too much here - this is only being used in test code so far, and probably won't get that much use in production either. Though in general, I think if I see a java object with a method called waitFor that takes an int argument timeout (which should probably be a long), then my immediate reaction is that I won't be waiting longer than the timeout specified. In this case, you could end up waiting 60 seconds with a 20 second timeout (in the worst case). If it's for 2 separate calls, then it should be specified in the documentation.

I can't remember where the good common implementation of async retry executers was. Maybe in apache or in guava. If I can't find it, it's not a lot of code to just have in iris common.

olaurino: I am changing the implementation using the Future API, so most of the issues you are highlighting will be gone.

olaurino:
Added some new commits.

The driver for the changes was to improve the retry logic and the timeout handling. This called for a refactoring and clean-up of the SherpaClient class, so that code duplication could be removed and more encapsulation could be achieved.

So, the waitFor method is gone, and there are instead two static factory methods that provide instances of the SAMPController and of the SherpaClient class.

Some improvements left for the future (we really need to integrate the PRs for this release now):
  * in the current API the asynchronous retry logic is hidden from the clients, who can only deal with synchronous calls. This is not a regression, but the clients might be exposed to the Future object and decide what to do.
  * Create factory classes instead of static factory methods (maybe with the AbstractFactory design pattern).
  * Exceptions handling can be greatly improved.

The Travis builds just finished and failed, even though all the tests passed on my system. I'll look into it.

olaurino:
The issue should now be fixed, and the Travis builds pass.

olaurino:
Closing this PR as it has been replaced by #207

PR #203, opened 2015-11-27
Iris 191 implement abstract test for it

jbudynk:
Address #191 

Depends on #200 (or another way to setup SAMP and Sherpa)

# Description
This PR checks that a SAMP Hub and Sherpa are up and running before any non-GUI integration tests that require a SAMP connection are executed.

All tests that require a SAMP connection are extended from `AbstractSAMPTest.java`. Before a test is run, an attempt is made to connect a `SedSAMPController` to a SAMP hub; if a connection cannot be made within 5 seconds, the test fails. Similarly, a test is run to make sure Sherpa is attached to the SAMP Hub. 

`AbstractSAMPTest.java` uses the `@Rule` annotation to setup/tear down tests, instead of using `@Before`/`@After`. Reasons for this change are that `@Rule` and JUnit's `ExternalResource` class may prove to be more flexible in test infrastructure down the road.

Note that the Travis builds are NOT passing, as **this PR depends on a way to have the SAMP hub and Sherpa up before the tests run**. However, the Travis test failures are a good thing - it shows that the tests fail because they can't connect to a SAMP hub (see Travis build [#203](https://travis-ci.org/ChandraCXC/iris/builds/93584102)). On my local machine, I've merged #200 onto these changes to make sure that the tests run and pass as expected. If necessary, I will make this branch dependent on #200; as a result of this, I will need to add `maven-antrun-plugin` to `iris-common/pom.xml`.

### DIFF
--- a/iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
+++ b/iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
@@ -20,9 +20,7 @@ import cfa.vo.iris.interop.SedSAMPController;
 import cfa.vo.sherpa.SherpaClient;
 import java.util.logging.Logger;
 import org.astrogrid.samp.client.SampException;
-import org.junit.After;
 import static org.junit.Assert.*;
-import org.junit.Before;
 import org.junit.Rule;
 import org.junit.rules.ExternalResource;
 
@@ -52,39 +50,29 @@ public class AbstractSAMPTest {
     }
     
     
-//    /**
-//     * This rule allows the component loader to be initialized only once for the whole suite.
-//     * This works around the fact that @BeforeClass and @AfterClass methods need to be static.
-//     */
-//    @Rule
-//    public ExternalResource resource = new ExternalResource() {
-//        @Override
-//        protected void before() throws Exception {
-//            // asserts that the SAMP Hub is up, and that sherpa-samp is connected.
-//	    connectToSAMPHub();
-//	    isSherpaConnected();
-//        }
-//        @Override
-//        protected void after() {
-//            // disconnect controller
-//	    controller.stop();
-//        }
-//    };
-    
-    // forces the concrete classes to make their own before class, and it will 
-    // not override this Before class
-    @Before
-    public final void setUp() throws Exception {
-    
-        // asserts that the SAMP Hub is up, and that sherpa-samp is connected.
-        connectToSAMPHub();
-        isSherpaConnected();
-    }
-    
-    @After
-    public void tearDown() {
-        controller.stop();
-    }
+    /**
+     * Create before/after methods for each test class that extends this class.
+     * Setup/teardown SedSAMPControllers after each test.
+     * 
+     * Note that in the future we can implement the ExternalResource interface
+     * for various kinds of tests (NED Service, SAMP, Sherpa, etc.) in 
+     * independent classes. Then a test that needs multiple setup/teardown 
+     * methods can just instantiate the necessary Rules.
+     */
+    @Rule
+    public ExternalResource resource = new ExternalResource() {
+        @Override
+        protected void before() throws Exception {
+            // asserts that the SAMP Hub is up, and that sherpa-samp is connected.
+            connectToSAMPHub();
+            isSherpaConnected();
+        }
+        @Override
+        protected void after() {
+            // disconnect controller
+            controller.stop();
+        }
+    };
     
     /*
     * Creates a SedSAMPController and connects it to a SAMP Hub. Fails if
--- a/iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
+++ b/iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
@@ -18,6 +18,7 @@ package cfa.vo.iris.test.unit.it;
 
 import cfa.vo.interop.SAMPController;
 import cfa.vo.sherpa.SherpaClient;
+import java.util.logging.Logger;
 import org.astrogrid.samp.client.SampException;
 import org.junit.After;
 import static org.junit.Assert.*;
@@ -31,21 +32,42 @@ import org.junit.Before;
  */
 public class AbstractSAMPTest {
     
+    private final Logger logger = Logger.getLogger(AbstractSAMPTest.class.getName());
+    
+    private final int SAMP_CONN_RETRIES = 3;
+    
     protected SAMPController controller;
     private int TIMEOUT;
     private String controller_name;
     protected SherpaClient client;
     
     public AbstractSAMPTest() {
-	this.TIMEOUT = 5;
-	this.controller_name = "Iris Component";
+	this.controller_name = "Iris Test Controller";
     }
     
     public AbstractSAMPTest(int timeout, String controller_name) {
-	this.TIMEOUT = timeout;
 	this.controller_name = controller_name;
     }
     
+    
+//    /**
+//     * This rule allows the component loader to be initialized only once for the whole suite.
+//     * This works around the fact that @BeforeClass and @AfterClass methods need to be static.
+//     */
+//    @Rule
+//    public ExternalResource resource = new ExternalResource() {
+//        @Override
+//        protected void before() throws Exception {
+//            // asserts that the SAMP Hub is up, and that sherpa-samp is connected.
+//	    connectToSAMPHub();
+//	    isSherpaConnected();
+//        }
+//        @Override
+//        protected void after() {
+//            // disconnect controller
+//	    controller.stop();
+//        }
+//    };
     
     // forces the concrete classes to make their own before class, and it will 
     // not override this Before class
@@ -76,17 +98,16 @@ public class AbstractSAMPTest {
 	
 	// If the controller doesn't connect after 2 seconds, add additional
 	// wait time.
-	int time = 0;
-	while (!controller.isConnected()) {
-	    time += TIMEOUT*1000;
-	    Thread.sleep(time);
-	    
-	    // If the controller doesn't connect within the additional time, 
-	    // fail.
-	    if (time >= TIMEOUT*1000) {
-		fail("Could not connect to SAMP Hub.");
-	    }
-	}
+	int count = 0;
+        while (!controller.isConnected()) {
+            if (++count > SAMP_CONN_RETRIES) {
+                String msg = "Failed to connect to SAMP, failing Unit tests";
+                logger.info(msg);
+                fail(msg);
+            }
+            logger.info("waiting connection");
+            Thread.sleep(1000);
+        }
 	assertTrue(controller.isConnected());
     }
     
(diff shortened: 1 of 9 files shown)


iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
```diff
@@ -0,0 +1,122 @@
...
+        @Override
+        protected void before() throws Exception {
+            // asserts that the SAMP Hub is up, and that sherpa-samp is connected.
+            connectToSAMPHub();
+            isSherpaConnected();
+        }
+        @Override
+        protected void after() {
+            // disconnect controller
+            controller.stop();
```
eholum: Is this the right behavior? Was removed in https://github.com/ChandraCXC/iris/commit/b9a4e95b2cf8caca4773bb936f75db9f40008496

iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /*
+    Connects a SedSAMPController to a SAMP Hub. Fails if it does not connect to a 
+    SAMP Hub.
+    */
+    public void connectToSAMPHub(SedSAMPController controller) throws InterruptedException, Exception {
+        controller.setAutoRunHub(false);
+        controller.start(false);
+        Thread.sleep(2000); // wait 2 seconds for connection
```
eholum: You can just let the loop below handle the waiting - it'll all be logged, then.

sed-builder/src/test/java/cfa/vo/sed/science/stacker/AbstracSEDStackerIT.java
```diff
@@ -2,26 +2,19 @@
...
-import cfa.vo.iris.interop.SedSAMPController;
+import cfa.vo.iris.test.unit.it.AbstractSAMPTest;
 
 /**
  * Abstract class for integration testing of SAMP integration. Tests will fail if they are
  * unable to connect to the SAMP hub.
  * 
  */
-public abstract class AbstracSEDStackerIT {
+public abstract class AbstracSEDStackerIT extends AbstractSAMPTest {
```
eholum: Blurgh, do you mind fixing the typo in the name of the test case?

iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
jbudynk: Oh nice catch. Will change and rerun the tests.

sed-builder/src/test/java/cfa/vo/sed/science/stacker/AbstracSEDStackerIT.java
jbudynk: Haha sure will do

iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
olaurino: Hmm... why shouldn't you stop the controller when tearing down the controller resource? b9a4e95 removed a left-over controller.stop() from a test method since stop() was already called in the tearDown method. Or am I missing something?

eholum: Ah, it's still in there,
https://github.com/ChandraCXC/iris/pull/202/files#diff-87b71316a96a127320bbb44aa1ada129R137


My mistake

On Tue, Dec 1, 2015 at 4:45 PM, Omar Laurino <notifications@github.com>
wrote:

> In
> iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
> <https://github.com/ChandraCXC/iris/pull/203#discussion_r46343805>:
>
> > +     * for various kinds of tests (NED Service, SAMP, Sherpa, etc.) in
> > +     * independent classes. Then a test that needs multiple setup/teardown
> > +     * methods can just instantiate the necessary Rules.
> > +     */
> > +    @Rule
> > +    public ExternalResource resource = new ExternalResource() {
> > +        @Override
> > +        protected void before() throws Exception {
> > +            // asserts that the SAMP Hub is up, and that sherpa-samp is connected.
> > +            connectToSAMPHub();
> > +            isSherpaConnected();
> > +        }
> > +        @Override
> > +        protected void after() {
> > +            // disconnect controller
> > +            controller.stop();
>
> Hmm... why shouldn't you stop the controller when tearing down the
> controller resource? b9a4e95
> <https://github.com/ChandraCXC/iris/commit/b9a4e95b2cf8caca4773bb936f75db9f40008496>
> removed a left-over controller.stop() from a test method since stop() was
> already called in the tearDown method. Or am I missing something?
>
> —
> <https://github.com/ChandraCXC/iris/pull/203/files#r46343805>.
>



-- 
Erik Holum

iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
jbudynk: Following your suggestion, on my local machine, I changed SAMP_CONN_RETRIES
= 5 and removed the extra Thread.sleep(1000) that was outside the loop.
You'd figure this would equate to the current connectToSAMPHub() logic, but
it doesn't. It's unstable; sometimes the tests would successfully run,
other times it took awhile for Sherpa to connect. We've experienced varying
sherpa-samp connection times in the past. We suggest to users if their
smoke tests are failing to provide a longer run time; this usually solves
the problem.

Indeed, if I increase SAMP_CONN_RETRIES to 10, I've had my tests
successfully run multiple times.

What's odd is that the tests always passed when I used commit 23beefd0469 (with
the addition of the maven-antrun-plugin in the iris-common POM).

On Tue, Dec 1, 2015 at 4:23 PM, Erik Holum <notifications@github.com> wrote:

> In
> iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
> <https://github.com/ChandraCXC/iris/pull/203#discussion_r46340894>:
>
> > +    controller = new SedSAMPController(
> > +            this.controller_name,
> > +            this.controller_name,
> > +            this.getClass().getResource("").toString()
> > +	);
> > +    connectToSAMPHub(controller);
> > +    }
> > +
> > +    /*
> > +    Connects a SedSAMPController to a SAMP Hub. Fails if it does not connect to a
> > +    SAMP Hub.
> > +    */
> > +    public void connectToSAMPHub(SedSAMPController controller) throws InterruptedException, Exception {
> > +        controller.setAutoRunHub(false);
> > +        controller.start(false);
> > +        Thread.sleep(2000); // wait 2 seconds for connection
>
> You can just let the loop below handle the waiting - it'll all be logged,
> then.
>
> —
> <https://github.com/ChandraCXC/iris/pull/203/files#r46340894>.
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

eholum: Yeah, it sounds like it can take a while for SAMP to get connect (more than a minute?). If it's just a matter of increasing the timeout (say to 30 or 40), it's probably worth it to make the test deterministic. This only has to be run once during the integration test phase, so it shouldn't add too much overhead.

olaurino: Also, in my experience this is more likely to be an issue on our
workstations rather than on travis. Not that I am saying this is good, it's
just what it is. On my workstation, where during development SAMP hubs are
started and shut down so many times, there are times when it may take up to
two minutes for a hub to show up, especially on Linux. Not sure whether
this is a sign of something going wrong, but it's hardly reproducible and
hard to track down.

Omar.

On Wed, Dec 2, 2015 at 9:15 AM, Erik Holum <notifications@github.com> wrote:

> In
> iris-common/src/test/java/cfa/vo/iris/test/unit/it/AbstractSAMPTest.java
> <https://github.com/ChandraCXC/iris/pull/203#discussion_r46418225>:
>
> > +    controller = new SedSAMPController(
> > +            this.controller_name,
> > +            this.controller_name,
> > +            this.getClass().getResource("").toString()
> > +	);
> > +    connectToSAMPHub(controller);
> > +    }
> > +
> > +    /*
> > +    Connects a SedSAMPController to a SAMP Hub. Fails if it does not connect to a
> > +    SAMP Hub.
> > +    */
> > +    public void connectToSAMPHub(SedSAMPController controller) throws InterruptedException, Exception {
> > +        controller.setAutoRunHub(false);
> > +        controller.start(false);
> > +        Thread.sleep(2000); // wait 2 seconds for connection
>
> Yeah, it sounds like it can take a while for SAMP to get connect (more
> than a minute?). If it's just a matter of increasing the timeout (say to 30
> or 40), it's probably worth it to make the test deterministic. This only
> has to be run once during the integration test phase, so it shouldn't add
> too much overhead.
>
> —
> <https://github.com/ChandraCXC/iris/pull/203/files#r46418225>.
>



-- 
Omar Laurino
02140 Cambridge, MA
(617) 495-7227

jbudynk: Ok, so the issue was there wasn't enough time for sherpa-samp to connect to the SAMP hub. I added `Thread.sleep(SAMP_CONN_RETRIES*1000)` before checking `assertTrue(!client.findSherpa().isEmpty())`. This is not ideal; I'd rather do something similar to the method in `connectToSAMPHub()` that keeps retrying the connection. I tried a method to do this (discussed in a comment), but it was causing an error in the science and sedstacker tests, so I reverted back to this simple method.

jbudynk:
So in commit 4505666, I had to do a dummy fix to make sure there's enough time for sherpa-samp to connect to the SAMP hub. As stated in the line comments, I make the tests wait `SAMP_CONN_RETRIES*1000` milliseconds (5 seconds) before checking if `SherpaClient().findSherpa().isEmpty()`. 

Obviously, this isn't ideal; I'd prefer to have a better retry method like in `connectToSAMPHub().` I tried doing the following to catch the exception raised when SherpaClient().findSherpa() fails:

```
void isSherpaConnected() {
        this.client = new SherpaClient(this.controller);
        boolean connected = false;
        int count = 0;
        while (!connected) {
            System.out.println("Went through "+count+" times.");
            if (++count > SAMP_CONN_RETRIES) {
                String msg = "Sherpa did not connect to SAMP hub, "
                        + "failing tests.";
                logger.info(msg);
                fail(msg);
            }
            try {
                this.client.findSherpa();
                connected = true;
            } catch (SampException ex) {
                // Wait sometime for Sherpa to connect to SAMP hub.
                logger.info("waiting connection");
                Thread.sleep(1000);
            }
        }
        assertTrue(!this.client.findSherpa().isEmpty());
```

This caused a `SampException` of "Synchronous call timeout" in the first test; the other tests failed with the JUnit `AssertionError`, "Sherpa did not connect to SAMP hub, failing tests."

Instead of looking into fixing these issues, since Omar is working on updates to the SherpaClient, I figured I'd revert these changes back to the simple method of waiting a few seconds before calling `findSherpa()`.

olaurino:
Closing this PR as it has been replaced by #207

PR #207, opened 2015-12-02
Integration of PRs #203 #202 #200 #197: integration test infrastructure

eholum:
Merging open PRs into a single branch. 

https://github.com/ChandraCXC/iris/pull/203
https://github.com/ChandraCXC/iris/pull/202
https://github.com/ChandraCXC/iris/pull/200
https://github.com/ChandraCXC/iris/pull/197

Branch off this for comments/suggestions for sprint II.

PR #208, opened 2015-12-04
refactor samp controller and introduce builder (also unignore test)

olaurino:
This PR adds some major refactoring to the samp controller creation mechanism to improve and streamline its use in Iris, especially to the benefit of the new testing infrastructure.

One test was still ignored in PR #207, and this PR unignores it as well.

Note that I got rid of the SedSAMPController and moved its only method to the ExtSed class.

I also started getting rid of some static methods of their usage, but more work will be needed there.

Unfortunately documentation (java docstrings) and unit testing of the new Builder for SAMPController is limited or non-existent. Hopefully I can add those today otherwise we'll leave it to a new task for the future.


iris-common/src/test/java/cfa/vo/iris/common/SedMessageIT.java
```diff
@@ -68,23 +66,20 @@ public void teardown() {
     public void sedMessageTest() throws Exception {
         System.setProperty("jsamp.hub.profiles", "std");
 
-        SedSAMPController sampReceiver = SedSAMPController.createAndStart("TestReceiver", 
-                "An SED builder from the Virtual Astronomical Observatory",
-                this.getClass().getResource("/iris_button_tiny.png"), 
-                false,
-                false);
+        long timeout = Default.getInstance().getSampTimeout().convertTo(TimeUnit.MILLISECONDS).getAmount();
+        SAMPController sampSender = new SAMPController.Builder("TestSender")
```
jbudynk: There is a controller created and connected to the SAMP hub before this test runs. In PR #203, I gave the protected `controller` (created in `AbstractSAMPTest.connectToSAMP()`) the same setup as `sampSender` in SedMessageIT; that way, there's only 2 SedSAMPControllers. Now, there will be three controllers connected to the SAMP hub. Is this ok?

sed-builder/src/test/java/cfa/vo/sed/science/interpolation/SherpaRedshifterIT.java
```diff
@@ -79,7 +78,7 @@ public void testShift() throws Exception {
         };
 
         // Start the SAMP controller
-        controller = SedSAMPController.createAndStart("SEDStacker", "SEDStacker", this.getClass().getResource("/tools_tiny.png"), false, false);
+        controller = new SAMPController.Builder("SEDStacker").buildAndStart(30000);
```
jbudynk: Following the changes in #203, this should be using the super class's `controller`; it should not be creating a new one. This was something I missed in #203.

iris-common/src/test/java/cfa/vo/iris/common/SedMessageIT.java
olaurino: Yes. This test needs a controller with a resource server, and the choice I made for the builder is that when you build the controller either you start it with a resource server or you don't.

The controller in the abstract class is not started with a resource server (and probably it should not be), so I need one more controller.

We can complicate the abstract class to be instructed on how to build the controller, but I don't think that's necessary. Extending classes can decide to use the parent's controller as it was created, or not. Just like they create new controllers if they need to test a SAMP conversation.

sed-builder/src/test/java/cfa/vo/sed/science/interpolation/SherpaRedshifterIT.java
olaurino: OK.

jbudynk:
Should test-components/src/main/java/cfa/vo/iris/test/TestSSAServer.java extend AbstractSAMPTest since it checks for a SAMP hub connection? Or at least should it be an IT test? I'm unfamiliar with this test component.

olaurino:
TestSSAServer is not a class in the test jar, it's in main, so it cannot access the test infrastructure. It is a genuine SAMP client used by the test builder (when running Iris with --test) to test that Iris correctly responds to a SAMP client sending SAMP messages, including ones with SSA (Simple Spectral Access Protocol) mtypes. The name is probably misleading because it started as an SSA server stub and then grew to be a larger utility components when testing release candidates.

sed-builder/src/test/java/cfa/vo/sed/science/interpolation/SherpaRedshifterIT.java
jbudynk: Alright, I'm making the changes and pushing them to #207.

olaurino:
Replaced by #228

PR #228, opened 2015-12-09
217 218 229 230 refactor samp controller

### LINKED ISSUE #217: SAMP Controller Refactor
Incl. Updating the status checker.

### LINKED ISSUE #218: Document New SAMP Controller Behavior

### LINKED ISSUE #229: Remove Thread.sleep from the BuilderMainViewTest
By putting something in that lets us wait for completion of the spawned Events' threads.

### LINKED ISSUE #230: Jacoco Build Issue
Fix an issue with Jacoco and incorrect/broken coverage analysis

olaurino:
Resolve #217
Resolve #218
Resolve #229
Resolve #230

This PR replaces #208.

This PR adds some major refactoring to the samp controller creation mechanism to improve and streamline its use in Iris, especially to the benefit of the new testing infrastructure, to which this PR contributes new Rule implementations.

One test was still ignored in PR #207, and this PR unignores it as well.

Note that I got rid of the SedSAMPController and moved its only method to the ExtSed class.

I also started getting rid of some static methods of their usage, but more work will be needed there.


samp-factory/src/main/java/cfa/vo/interop/SAMPController.java
```diff
@@ -212,53 +220,43 @@ public void sendMessage(SAMPMessage message, ResultHandler handler, int timeout)
...
+                    Thread.sleep(1000); // This will be interrupted if a timeout occurs
+                }
+                return Boolean.TRUE;
+            }
+        };
+        Future<Boolean> futureController = executor.submit(callable);
+        try {
+            return futureController.get(timeoutMillis, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException | ExecutionException | TimeoutException ex) {
+            futureController.cancel(true);
```
eholum: Finally shut down the executor

iris-common/src/main/java/cfa/vo/interop/HubSAMPController.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        private static HubSAMPController getInstance() {
+            return SingletonHolder.INSTANCE;
+        }
+    }
+
+    public static HubSAMPController getInstance(SAMPControllerBuilder builder, long timeoutMillis) throws Exception {
+        if (HubSAMPController.builder != null) {
+            throw new IllegalStateException("Class was already instantiated with a different argument");
```
eholum: Not have it return the instance?

iris-common/src/main/java/cfa/vo/interop/HubSAMPController.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        HubSAMPController.timeoutMillis = timeoutMillis;
+        HubSAMPController.builder = builder;
+
+        try {
+            return SingletonHolder.getInstance();
+        } catch (ExceptionInInitializerError ex) {
```
eholum: Don't wrap the exception above in the initHubController method, then just let the exception thrown there pass through.

samp-factory/src/main/java/cfa/vo/interop/SAMPController.java
olaurino: Yes, this was fixed in a commit I haven't published yet.

iris-common/src/main/java/cfa/vo/interop/HubSAMPController.java
olaurino: I thought what's missing is a `getInstance()` method without arguments, but that has its issues too. I just don't think that a singleton with arguments really works.

In the commits I have still unpublished I am making this class a private class inside `AbstractIrisApplication`. This is the only class that needs it. At that point, probably, you don't even need to bother if it's a singleton.

iris-common/src/main/java/cfa/vo/interop/HubSAMPController.java
olaurino: As I mentioned above the whole "singleton with arguments" thing doesn't convince me and this code is likely to change anyway.

iris-common/src/main/java/cfa/vo/iris/AbstractIrisApplication.java
```diff
@@ -276,4 +282,95 @@ public void addMessageHandler(MessageHandler handler) {
...
+            if (hub != null) {
+                hub.shutdown();
+            }
+            super.stop();
+        }
+
+        @Override
+        public boolean start(long timeoutMillis) {
+            if(!started) {
+                timer = new Timer(true);
```
eholum: The downside to a timer is that it starts and maintains a single background thread for the duration of it's life. Any exceptions or issues with that thread will kill the timer - we could be seeing an issue with autorun hub dying and killing the retry. A ScheduledThreadPoolExecutor will spawn a new thread as needed - and will generally be more robust in terms of the Iris application's life. Since it's easy enough to modify the CheckConnectionTask to be a Runnable, I think it would be preferable to use an Executor as opposed to a Timer.

iris-common/src/main/java/cfa/vo/iris/AbstractIrisApplication.java
```diff
@@ -195,9 +192,17 @@ protected void initComponents() {
...
+                        .getSampTimeout()
+                        .convertTo(TimeUnit.MILLISECONDS)
+                        .getAmount();
+                SAMPControllerBuilder builder = new SAMPControllerBuilder(getName())
+                        .withDescription(getDescription())
+                        .withResourceServer("sedImporter/")
+                        .withIcon(getSAMPIcon());
+                sampController = new HubSAMPController(builder, timeout);
+
             } catch (Exception ex) {
```
eholum: Should this be an application killing exception? If there's an exception here will the Iris application ever be able to connect to a SAMP hub?

iris-common/src/main/java/cfa/vo/iris/sed/ExtSed.java
```diff
@@ -310,4 +316,46 @@ public static ExtSed flatten(ExtSed sed, String xunit, String yunit) throws cfa.
...
+                msg.setUrl(url.toString());
+
+                SAMPMessage message = SAMPFactory.createMessage("table.load.votable", msg, VaoMessage.class);
+                ((Map)message.get().getParam("vao-payload")).put("sed-id", getId());
+                controller.sendMessage(message);
+            }
+
+        } catch (Exception ex) {
+            Logger.getLogger(ExtSed.class.getName()).log(Level.SEVERE, null, ex);
+            throw new SampException(ex.getMessage());
```
eholum: Just wrap ex so that we have the full stack trace.

iris-common/src/main/java/cfa/vo/iris/AbstractIrisApplication.java
```diff
@@ -276,4 +282,95 @@ public void addMessageHandler(MessageHandler handler) {
...
+        public void setAutoRunHub(boolean autoRunHub) {
+            this.autoRunHub = autoRunHub;
+        }
+
+        private class CheckConnectionTask extends TimerTask {
+
+            private boolean state = false;
+
+            private void runHubIfNeeded() {
+                // I autorunhub is false we don't want to start a new hub.
```
eholum: mt

iris-common/src/main/java/cfa/vo/iris/AbstractIrisApplication.java
```diff
@@ -276,4 +282,95 @@ public void addMessageHandler(MessageHandler handler) {
...
+                    // if we are connected, or if the hub is null and autoRunHub is false, then there is nothing to do.
+                    // Keep monitoring
+                } catch (IOException ex) {
+                    // do nothing, keep monitoring
+                    logger.log(Level.INFO, "a hub is already running? Moving on", ex);
+                }
+            }
+
+            private void checkStatusUpdate() {
+                boolean stateChanged = state != isConnected();
```
eholum: Log the current state - maybe at debug level

iris-common/src/main/java/cfa/vo/sherpa/SherpaClient.java
```diff
@@ -233,11 +237,13 @@ public SherpaClient call() throws Exception {
                 return client;
             }
         };
-        Future<SherpaClient> futureSherpaClient = pool.submit(callable);
+        Future<SherpaClient> futureSherpaClient = executor.submit(callable);
         try {
             return futureSherpaClient.get(timeout.getAmount(), timeout.getUnit());
         } catch (Exception ex) {
             throw new RuntimeException("Cannot find Sherpa!");
```
eholum: Wrap the exception here, we're losing information in the tests (always generally good practice).

samp-factory/src/main/java/cfa/vo/interop/SAMPController.java
```diff
@@ -212,53 +215,40 @@ public void sendMessage(SAMPMessage message, ResultHandler handler, int timeout)
...
+                    Thread.sleep(1000); // This will be interrupted if a timeout occurs
+                }
+                return Boolean.TRUE;
             }
+        };
+        Future<Boolean> futureController = executor.submit(callable);
+        try {
+            return futureController.get(timeoutMillis, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException | ExecutionException | TimeoutException ex) {
+            futureController.cancel(true);
```
eholum: Log the exception

olaurino:
Replaced by #231.

PR #231, opened 2016-01-13
217 218 229 230 Big refactoring and cleanup

### LINKED ISSUE #217: SAMP Controller Refactor

### LINKED ISSUE #218: Document New SAMP Controller Behavior

### LINKED ISSUE #229: Remove Thread.sleep from the BuilderMainViewTest

### LINKED ISSUE #230: Jacoco Build Issue

olaurino:
This PR builds on previous PRs/branches and introduces refactorings and fixes to make sure the SAMP testing issues are fixed (at least most of the time).

Several changes are embedded in the updates to sherpa-samp.

Replaces #228
Resolve #217
Resolve #218
Resolve #229 
Resolve #230

### DIFF
--- a/iris/src/main/java/cfa/vo/iris/Iris.java
+++ b/iris/src/main/java/cfa/vo/iris/Iris.java
@@ -122,7 +122,8 @@ public class Iris extends Application implements IrisApplication {
     protected void initialize(String[] args) {
         commandLine = new CommandLine(args);
         new Configuration(commandLine).apply();
-        componentLoader = new ComponentLoader(getClass().getResource("/components"));
+        componentLoader = new ComponentLoader(getClass().getResource("/components"),
+                                              commandLine.getAdditionalComponents());
         sampSetup();
     }
 
--- a/iris-common/src/main/java/cfa/vo/iris/ComponentLoader.java
+++ b/iris-common/src/main/java/cfa/vo/iris/ComponentLoader.java
@@ -32,18 +32,28 @@ public class ComponentLoader {
     protected List<String> failures = new ArrayList<>();
 
     public ComponentLoader(Collection<Class<? extends IrisComponent>> componentList) {
-        for (Class c : componentList) {
-            loadComponent(c);
-        }
+        addComponents(componentList);
     }
     
-    public ComponentLoader(URL componentsURL) {
+    public ComponentLoader(URL componentsURL, Collection<Class<? extends IrisComponent>> componentList) {
         try {
             initComponents(readComponentsFile(componentsURL));
         } catch (IOException ex) {
             String message = "Cannot read components file at: " + componentsURL;
             System.err.println(message);
             Logger.getLogger(ComponentLoader.class.getName()).log(Level.SEVERE, message, ex);
+        }
+        
+        addComponents(componentList);
+    }
+    
+    private void addComponents(Collection<Class<? extends IrisComponent>> componentList) {
+        if (componentList == null) {
+            return;
+        }
+        
+        for (Class<? extends IrisComponent> c : componentList) {
+            loadComponent(c);
         }
     }
 
--- a/iris-common/src/test/java/cfa/vo/iris/ComponentLoaderTest.java
+++ b/iris-common/src/test/java/cfa/vo/iris/ComponentLoaderTest.java
@@ -40,7 +40,7 @@ public class ComponentLoaderTest {
 
     @Test
     public void testLoadComponents() {
-        ComponentLoader loader = new ComponentLoader(testUrl);
+        ComponentLoader loader = new ComponentLoader(testUrl, new ArrayList());
         Collection<IrisComponent> components = loader.getComponents();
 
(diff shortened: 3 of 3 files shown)


jbudynk:
I'm still reviewing the code, but a a few general things I noticed while testing out the branch:

1.  On OS X, once in a while `mvn clean verify` or `mvn clean package` will fail at RedshiftInterpolateIT. 2/8 times of running either `mvn clean verify` or `mvn clean package` this happened.

2.  On Linux and OS X: If sherpa gets disconnected from the SAMP hub, the Sherpa-samp connection icon stays "connected." It only disconnects when Iris disconnects from the SAMP hub. I tried using SedStacker when I knew sherpa was disconnected, and the operations failed as expected.

jbudynk:
Command line arguments "`iris --test`" and "`iris --vizier`" no longer bring up a desktop icon for me (on OS X). I've run `mvn package` and `mvn install`; neither build brings up a TestBuilder or VizieR client on the desktop. 

I think the issue is in "Iris.java::initialize()" at [lines 122-125](https://github.com/ChandraCXC/iris/pull/231/files#diff-30fe77a8dbe4f4ef4d901626e96fe337R122). It doesn't look like CommandLine components get loaded with the ComponentLoader or with Configuration().apply().

iris/src/main/java/cfa/vo/iris/Iris.java
```diff
@@ -15,35 +15,71 @@
...
 import javax.swing.JDialog;
-import javax.swing.UIManager;
 
 /**
  * The main class of the application.
  */
-public class Iris extends AbstractIrisApplication {
+public class Iris extends Application implements IrisApplication {
+
+    private final Logger logger = Logger.getLogger(Iris.class.getName());
```
eholum: static

samp-factory/src/main/java/cfa/vo/interop/SAMPServiceBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Set whether the controller should start and internal HTTP server
+     * @param serverRoot the root path of the resource server
+     * @return
+     */
+    public SAMPServiceBuilder withResourceServer(String serverRoot) {
+        this.withResourceServer = true;
+        if (serverRoot != null) {
```
eholum: This is set above, so the serverRoot will never actually change?

samp-factory/src/main/java/cfa/vo/interop/SampService.java
```diff
@@ -0,0 +1,317 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class SampService {
+    private final String PING_MTYPE = "samp.app.ping";
```
eholum: statics

samp-factory/src/main/java/cfa/vo/interop/SampService.java
```diff
@@ -0,0 +1,317 @@
...
+        started = false;
+    }
+
+    public void addSampConnectionListener(SAMPConnectionListener listener) {
+        sampListeners.add(listener);
+        listener.run(sampUp);
+    }
+
+    private void monitorSampOnce() {
+        logger.log(Level.INFO, "Monitor State: ");
```
eholum: I might put just one line at the INFO level, and these other ones at debug, this is a lot of info to dump out every second.

samp-factory/src/main/java/cfa/vo/sherpa/SherpaClient.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Logger;
+
+public class SherpaClient {
+
+    private SampService sampService;
+    private Map<String, AbstractModel> modelMap = new HashMap<>();
+    private Integer stringCounter = 0;
+    private Logger logger = Logger.getLogger(SherpaClient.class.getName());
```
eholum: static

olaurino:
As we agreed, I'll merge this PR. Some issues are still open, but we have more tasks for them:
  * If sherpa disconnects, the desktop icon still shows it connected (#233).
  * Some components (we know about SedStacker for sure) freeze the GUI during a sherpa call. With the new retry code in place, this means that the GUI is frozen for 10 seconds before Iris gives up. We need to a) make sure those components do not freeze the GUI, and b) the user is aware that something is wrong but we are retrying. (#234)
  * Erratically, some tests will fail. So far we have seen that happening with the Redshift Integration Test, and we may want to investigate some more (#235).

There was also an issue with some command line arguments being ignored, but that's fixed by #241, which was already incorporated in this PR.

Issue #225, opened 2015-12-07
Review Semantics of SED in Segment Events


eholum:

eholum:
Moved to https://github.com/ChandraCXC/iris-dev/issues/48

Issue #236, opened 2016-01-19
Add Unit Tests for the Plotter


eholum:

eholum:
This issue was moved to ChandraCXC/iris-dev#23

PR #244, opened 2016-01-21
subscribe plotter to events, unit tests

### LINKED ISSUE #224: Subscribe Plotter to Event Listeners

olaurino:
Resolve #224 

# Description

This PR subscribes the plotter to sed events. All sed events trigger a reset of the plotter view. This may or may not be the desired eventual behavior, depending also on the work in #225.

For instance, the `reset` logic does not discriminate between resetting the viewport after a click of the "reset" button or plotting a new SED once it has been selected. Also, one may want to attach the viewport features as SED attachments and then restore the customized viewport once an SED regains focus, so that user's customization are preserved, at least during a session.

So, we may want to split `reset` into two methods with different scopes. This seems out of the scope of this PR, but we may decide otherwise.

There are many more unit tests that this PR may add, but it is hard to implement them without basic unit tests for the plotter itself, and much of the state of the plotter seems rather opaque at this point.


jbudynk:
Looks good to me. I agree that we should have two `reset` methods as you suggest.

I would say 'merge it' if we're OK with the single unit test added here. But if we want to wait 'till #236 is fixed and merged, that's fine.

PR #245, opened 2016-01-21
Iris 239 plotter performance test

### LINKED ISSUE #239: Add Performance Tests for the Plotter
Use SED builder to load 300k point set, the plot it

olaurino:
Resolve #239.

# Description
This PR introduces a simple benchmark assessing SEDs can be loaded and displayed in a reasonable amount of time.

This PR relies on a new git repository for the test data, so to reduce the footprint of the main repository.

The benchmark relies on a simple annotation from JUnit which will fail the test if it takes longer than a timeout. The timeout keeps into account the test setup time, so some heuristics were used to determine a reasonable timeout of 2500 milliseconds. Interactive testing was performed to ensure that the system is responsive.

We can use something fancier or try to roll our own benchmark logic if required.

I also had to remove some code that was added to warn the users when too large files were loaded. 

Plus, since SEDLib does not handle URLs directly, I overrode the `read` method in `ExtSed` to make the tests more portable (they were failing depending on the way they were called).

# Notes
The test is not particularly safe because there is currently no way to test whether the plot was actually displayed.

# Dependencies
This PR depends on #224 because I needed at least a naive way of checking that the plotter updated itself to plot the new sed. So, this PR should be rebased and merged after #224.

### DIFF
--- a/iris-visualizer/src/test/java/cfa/vo/iris/visualizer/VisualizerComponentTest.java
+++ b/iris-visualizer/src/test/java/cfa/vo/iris/visualizer/VisualizerComponentTest.java
@@ -73,7 +73,8 @@ public class VisualizerComponentTest extends AbstractComponentGUITest {
                 .getSubMenu(windowName)
                 .getSubMenu(windowName)
                 .click();
-        assertSame(sedManager.getSelected(), comp.getDefaultPlotterView().getSed());
+        ExtSed initialSed = sedManager.getSelected();
+        assertNull(initialSed);
 
         // Test the plotter reacts to a sed event
         final ExtSed sed = sedManager.newSed("sampleSed");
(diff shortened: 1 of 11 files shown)


iris-visualizer/src/test/java/cfa/vo/iris/visualizer/VisualizerComponentTest.java
```diff
@@ -52,4 +61,104 @@ public void testVisualizerStub() throws Exception {
...
+            public void run() {
+                assertSame(sed, comp.getDefaultPlotterView().getSed());
+            }
+        });
+
+
+//        assertEquals("sampleSed", comp.getDefaultPlotterView().getLegend().getTitle());
+    }
+
+    @Test(timeout=2500)
```
eholum: This junit timeout is sufficient for now. I like the look of the jmh, but don't think we'll need any heavy performance infrastructure yet, as we're not enforcing any strict SLAs on our functions.

Might be worth moving this to it's own integration tests though, if it does end up timing out from time to time we can just run it with the IT/functional tests.

eholum: Though it runs really quickly on my desktop, so probably fine where it is.

eholum:
I like the looks of both https://github.com/ChandraCXC/iris/pull/244 and this PR.

Yeah, the reset logic needs to be re-worked and broken down. Right now it's doing two jobs that aren't really supposed to be related - resetting the view of the plot and resetting the selected sed.

So I say merge it.

jbudynk:
I'm glad we can get rid of the ugly spectralWarning code now :) Good catch on removing it.

The `ExtSed.read()`updates are a nice addition too.

Both PRs look good to me. I agree with you and Erik - we should break down the `reset` functionality into two methods.

Concerning JMH and JUnit for performance tests: As long as all we care about is getting the tests to pass under a certain amount of time, I'd say using JUnit's timeout parameter is fine.

To think of other performance tests off the top of my head (in case anyone can think of a reason we'd want a more in-depth performance framework like JMH):

   1.  dealing with 1,000,000 points on the plotter, be it
       1. one segment, 
       2. multiple segments in one SED, or
       3. co-plotted SEDs
   2.  SED Stacker timing
   3.  Fitting:
       1. Calculating confidence intervals
   4.  Calculating fluxes under the model

Issue #238, opened 2016-01-19
Add Support for viewing individual segments to the metadata browser


eholum:

eholum:
NR

PR #246, opened 2016-01-22
Metadata Browser Prototype Implementation

eholum:
Diffing this against https://github.com/ChandraCXC/iris/pull/245 since I pulled in Omar's changes and worked mine over them.

Closes https://github.com/ChandraCXC/iris/issues/237 and https://github.com/ChandraCXC/iris/issues/238

In this PR:
* A functional version of the metadata browser that lists every segment in the SED and displays the star table when a segment is selected.
* Unit tests for verifying the above statements using a combination of window clicking through uispec4j and plain old java verification
* I moved test segment creation up one layer to the abstract class, also was having trouble getting the perf test passing on my box (1 in 5 times), so i just upped the timeout.
* Added a map to the segment star table map adapter class so as to not put duplicate star tables out there for things that are already in use.

PR #247, opened 2016-01-22
Basic implementation of the Iris Plotter

olaurino:
This PR will have additional unit tests added as apart of https://github.com/ChandraCXC/iris/issues/239 and https://github.com/ChandraCXC/iris/issues/236. As well as a cleaned up and more functional metadatabrowser prototype as apart of https://github.com/ChandraCXC/iris/issues/238.

So while there is still quite a bit of work to do, what we have here unblocks further development on the plotter and metadata browser by providing:

* A very basic implementation of a star table adapter class and a wrapper segmentstartable, giving the visualizer access to the basic relevant information in the selected SED.
* A plotter that leverages the adapater class to plot the selected SED's x, y, and error bars.
* A basic implementation of plot/layer preferences and a map that manages them, more enums will need to be added as we implement support for user-set preferences.


olaurino:
I quickly reviewed this PR at 89abb25. It looks good to me. If I had any comments they would be in parts of the code we have tasks for anyway, so let's just merge this and start from there (or rather, from rebasing #246.

olaurino:
Scratch that. We should merge #246 into this one and then rebase our work off the new master. Let me just make sure I don't have any significant comments on #246.

PR #248, opened 2016-01-22
Metadata browser impl #238

eholum:
Okay, rebased against iris-visualizer-impl.

Closes https://github.com/ChandraCXC/iris/issues/237 and https://github.com/ChandraCXC/iris/issues/238

In this PR:
* A functional version of the metadata browser that lists every segment in the SED and displays the star table when a segment is selected.
* Unit tests for verifying the above statements using a combination of window clicking through uispec4j and plain old java verification
* I moved test segment creation up one layer to the abstract class, also was having trouble getting the perf test passing on my box (1 in 5 times), so i just upped the timeout.
* Added a map to the segment star table map adapter class so as to not put duplicate star tables out there for things that are already in use.


eholum:
I pushed a change to use invokeAndWait in our GUI tests, which had 2 unfortunate side effects:

1) The performance tests for the large SED timing went up to 10 seconds on average, so I bumped the timeout up and moved it to an integration test,
2) https://travis-ci.org/ChandraCXC/iris/jobs/104178197#L798 - which doesn't happen when I run the test in eclipse nor on the cli, so I'm working on recreating it and understanding what is going wrong with stil. 

https://github.com/Starlink/starjava/blob/master/ttools/src/main/uk/ac/starlink/ttools/plot2/data/SmartColumnFactory.java#L182

As it's not immediately obvious to me what the plotter is attempting to do here - i'll keep looking.

olaurino:
The error in the build is generated when the assertion is run before the
plotter has updated itself with the new data, or at least that is what I
decided after running the test in the debugger and getting it to pass
without any changes.

Which is why I wrapped the assertion in a runnable and enqueued it in the
EDT. However, it is indeed possible that my workaround simply moved the
dust under the rug, since as you pointed out the exception is then thrown
in the EDT itself.

On Fri, Jan 22, 2016 at 4:04 PM, Erik Holum <notifications@github.com>
wrote:

> I pushed a change to use invokeAndWait in our GUI tests, which had 2
> unfortunate side effects:
>
> 1) The performance tests for the large SED timing went up to 10 seconds on
> average, so I bumped the timeout up and moved it to an integration test,
> 2) https://travis-ci.org/ChandraCXC/iris/jobs/104178197#L798 - which
> doesn't happen when I run the test in eclipse nor on the cli, so I'm
> working on recreating it and understanding what is going wrong with stil.
>
>
> https://github.com/Starlink/starjava/blob/master/ttools/src/main/uk/ac/starlink/ttools/plot2/data/SmartColumnFactory.java#L182
>
> As it's not immediately obvious to me what the plotter is attempting to do
> here - i'll keep looking.
>
> —
> <https://github.com/ChandraCXC/iris/pull/248#issuecomment-174047014>.
>



-- 
Omar Laurino
02140 Cambridge, MA
(617) 495-7227

eholum:
Okay, updated this with a change for verifying some of the Swing unit tests - passes on my desktop and in eclipse, still waiting on the Travis builds to finish.

iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
```diff
@@ -47,11 +47,9 @@ public SegmentStarTable(ISegment data) {
...
         this.columnInfo = createColumnInfo();
         
-        if (data.getDataID() != null && 
-            data.getDataID().getTitle() != null &&
-            !StringUtils.isEmpty(data.getDataID().getTitle().getId())) 
-        {
-            this.id = data.getDataID().getTitle().getId();
+        if (data.isSetTarget() &&
+            data.getTarget().isSetName()) {
+            this.id = data.getTarget().getName().getValue();
```
olaurino: What if we had a utility class that worked something like this:
~~~~ java
this.id = (String) SedlibGetter.get(instance, 'target.name.value', default_value)
~~~~

All the checks would be done in that method, and the code would be much cleaner.

Drawbacks would be inefficiency introduced by reflection and potential errors introduced by using strings, including type unsafety, but that might be mitigated by:

~~~~ java
this.id = (String) SedlibGetter.get(instance, 'target.name.value', default_value, String.class)
~~~~

Still type unsafe, but it might capture obvious issues.

I am really not sure what's worse.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/MetadataBrowserViewTest.java
```diff
@@ -0,0 +1,130 @@
...
+        mbView = comp.getDefaultPlotterView().getMetadataBrowserView();
+    }
+
+    @Test
+    public void testMetadataBrowser() throws Exception {
+        
+        // Load an sed into the workspace and ensure it shows up in the MB
+        final ExtSed sed = sedManager.newSed("test1");
+        sedManager.select(sed);
+        mbView.reset();
```
olaurino: Are we missing a task for subscribing the metadata browser to events?

eholum:
Yeah, ultimately we're going to need something like what's in the SEDBuilder to label and assign unique ids to segments, but I'm not sure what the correct answer is yet. Perhaps with what's coming in changes w.r.t stil integration the id issue will be more clear. Right now this was just an attempt to get a reasonable name out of the segment to display in the visualizer.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/MetadataBrowserViewTest.java
eholum: Right now the MB is tied to the visualizer through the plotter, so it's being updated at the same time as the plotter. If we want them to be separate we can add a task for it.

olaurino: So why do we need to explicitly reset the metadata browser here?

On Wed, Jan 27, 2016 at 11:29 AM, Erik Holum <notifications@github.com>
wrote:

> In
> iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/MetadataBrowserViewTest.java
> <https://github.com/ChandraCXC/iris/pull/248#discussion_r51009275>:
>
> > +        org.uispec4j.Button mbButton = desktop.getWindow(plWindowName).getButton("Metadata");
> > +        mbButton.click();
> > +
> > +        desktop.containsWindow(mbWindowName).check();
> > +        mbWindow = desktop.getWindow(mbWindowName);
> > +
> > +        mbView = comp.getDefaultPlotterView().getMetadataBrowserView();
> > +    }
> > +
> > +    @Test
> > +    public void testMetadataBrowser() throws Exception {
> > +
> > +        // Load an sed into the workspace and ensure it shows up in the MB
> > +        final ExtSed sed = sedManager.newSed("test1");
> > +        sedManager.select(sed);
> > +        mbView.reset();
>
> Right now the MB is tied to the visualizer through the plotter, so it's
> being updated at the same time as the plotter. If we want them to be
> separate we can add a task for it.
>
> —
> <https://github.com/ChandraCXC/iris/pull/248/files#r51009275>.
>



-- 
Omar Laurino
02140 Cambridge, MA
(617) 495-7227

eholum: We don't, I wasn't explicitly testing the subscription here - so just manually called reset to ensure it was set before verification. I can remove it and put the asserts into the swing executor queue if that would be your preference.

olaurino: This way we avoid the complexity of executing assertions in the EDT, so I am fine with it. As long as we test that the subscription works somewhere else. I couldn't find such test, though. Let's merge this PR, then if we need to add a test let's do it (or, if I missed it, just point me to it ;).

eholum: I'll add a test for it. We don't explicitly test the MB reset in the other visualizer test anywhere else, so we should have it here.

PR #251, opened 2016-01-27
Add event listener verification for the Metadata Browser.

eholum:
Adding a unit test for the metadata browser for verifying that changes in the SedManager are correctly received and handled by the MB.


olaurino:
Travis builds are green, merging.

PR #252, opened 2016-01-28
iris-dev#20 Flatten StarTable

olaurino:
Resolve ChandraCXC/iris-dev#20
Resolve ChandraCXC/iris-dev#6

This PR introduces methods to _flatten_ a number of StarTables into a single StarTable, i.e. to provide a unified view of a number of tables. This is both a concatenation (values in common columns are concatenated) and a join (columns unique to one or more tables are null-padded and attached to the output table).

Columns are "common" in a semantic sense: they may have different names and units, for instance, but if the utype is the same, then they are considered the same and concatenated. Also, if the units differ, the values are converted to the same unit in the output table.

The current implementation is incomplete and focuses on columns. It is also rather inefficient, because it copies all the values over, but I want to make sure I have all the tests in place and a simple working implementation before I optimize it. Eventually, I would like to just store original tables in the flattened "view" and implement the StarTable interface with some form of smart delegation. I still don't know how to avoid potential memory leaks, but maybe that must be dealt with by the calling code once we integrate this. It is also hard to make sure the implementation is robust against changes in the shape of the tables, or in the column metadata. Again, this is possibly something to be taken into account in the client code, i.e. if disruptive changes are made, the old view should be discarded and a new one created. This is one more reason to make the view creation as efficient as possible.

If two tables have columns with the same ID, the output table will have a column with the same ID. Values will be concatenated, maintaining the order in which the tables are passed as arguments.

If a table has a column with a unique ID among the set of tables being concatenated, then a new column is created in the output table with all nulls except the values of the table containing that column.

IDs for columns are determined by taking the utype if one exists. Otherwise, ColumnInfo names will be used. More complicated rules can be applied if necessary.

Note that this PR also addresses ChandraCXC/iris-dev#6, as I needed such a method in my implementation.

Still to be done:
  * homogenize units
  * what about columns with same IDs but different UCDs?
  * copy parameters over (but what if two tables have parameters with same ID and different value?)
  * ~~compare utypes by matching rather than by equality (e.g. `ssa:Target.name` and `spec:spectrum.target.name` are the same)~~
  * optimize implementation
  * are nulls OK? Probably STIL provides utility methods to determine the best null value for each content class
  * error handling:
    * incompatible columns with same ID
    * unsupported unit strings
  * refactoring
  * integration in metadata browser and plotter

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/sed/stil/StilColumnManager.java
+++ b/iris-common/src/main/java/cfa/vo/iris/sed/stil/StilColumnManager.java
@@ -207,6 +207,7 @@ public class StilColumnManager {
 
             for (ColumnInfo info : index.getValues()) {
                 List<Object> arrayList = new ArrayList<>();
+                Object fillValue = info.getContentClass() == Double.class? Double.NaN : null;
                 tableLoop:
                 for (StarTable t : tables) {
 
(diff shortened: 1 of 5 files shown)


olaurino:
I guess this PR should be closed, right @eholum @jbudynk?

jbudynk:
makes sense - close it

On Tue, Apr 26, 2016 at 9:45 AM, Omar Laurino <notifications@github.com>
wrote:

> I guess this PR should be closed, right @eholum
> <https://github.com/eholum> @jbudynk <https://github.com/jbudynk>?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/ChandraCXC/iris/pull/252#issuecomment-214750025>
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

PR #255, opened 2016-02-08
Preferences Infrastructure for Iris Visualizer

eholum:
Closes https://github.com/ChandraCXC/iris-dev/issues/8

This is an implementation for preferences/settings for the iris plotter. Preferences are in two beans - PlotPreferences and SegmentLayers, which manage settings for the plot and each segment in the currently selected SED.

I moved management of the settings into a single location - VisualizerComponentPreferences, which handles creating, updating, retrieving, and notifying any listeners to changes in the preferences or in the SED workspace by leveraging it's own change notification and listener classes. The idea here is that different changes will allow the plotter and the MB to reset, redraw, or ignore changes that might not be relevant at the time independently from what's happening in the SED manager. It also ensures that any serialization or changes occurring in the workspace are handled prior to displaying any data, and that serialization only happens once per change - but there are additional time optimizations that we can add pending an implementation of column matchers.

I've included settings for most of the options available in Stil, but it meets the current requirements laid out in the visualizer stories, and is at least easily extensible to include any of them.

Also, since our existing swing tests rely on async processes in the SwingEventQueue, I found it necessary to add retry verification for assertions that for formally just single attempts.


jbudynk:
minor issue: There are a number of files introduced in this PR that don't have license headers. Could you add those in?

eholum:
Whoops, yep, added. Sorry about that.

PR #257, opened 2016-02-12
Generic Segment StarTable for Plotting and Showing Metadata

eholum:
This is kind of still a work in progress, in that there are many places that need optimization so that everything works more smoothly with large data sets.

The base class here is the SegmentStarTable refactor - which now, instead of just wrapping a segment, copies data into between 2 and 6 double data columns for flux and spectral data and error bounds. It also includes methods for replacing values in those columns and handling units as necessary. For the Visualizer component, the SegmentStarTable is joined to the serialized star table (using the vo serializer from sedlib) in an IrisStarTable, which lets us see both the plotted data and the original data in place.

This has made the plotter performance test ridiculously slow, which is because of the resetting, the unit stuff, and the vo serialization. That being said, things still load into iris, but we really need to finish https://github.com/ChandraCXC/iris-dev/issues/7 so that the plotter doesn't re-make every star table every time we want to add a segment. I'm working on that now, but wanted to put this out there to see if we're on the right track with it.

Long story short, this would allow plotting the correct values in the correct units - as specified in https://github.com/ChandraCXC/iris-dev/issues/6, but leave it open to replacing the plotter values with whatever other columns we wanted by just directly passing double arrays to the segmentstartable.


jbudynk:
minor issue: copyright headers missing from a few files

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/SegmentLayer.java
```diff
@@ -287,11 +338,14 @@ public SegmentLayer setErrorColorWeight(double errorColorWeight) {
...
+     * TODO: These setter methods should be made public when we support plotting arbitrary columns.
+     */
     public String getxCol() {
         return xCol;
     }
 
-    public SegmentLayer setxCol(String xCol) {
+    private SegmentLayer setxCol(String xCol) {
         this.xCol = xCol;
         return this;
```
jbudynk: Do all of these setter methods need return the SegmentLayer?

eholum: Just a style choice, makes it cleaner to chain setter calls together as in the constructor or where ever they're used. I don't think method chaining is always the best, but in the case of builders for java beans it's nice to have.

eholum:
Ah, I added the copyright config on my linux machine but not my mac. I'll add the missing ones, too.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/SegmentLayer.java
```diff
@@ -67,25 +73,18 @@
...
-            .setyErrHi(ColumnName.Y_ERR_HI.name())
-            .setyErrLo(ColumnName.Y_ERR_LO.name())
-            
-            // Setting default values here
+        // Setting default values here
+        this.setxCol(Column.SPECTRAL_COL.name())
+            .setyCol(Column.FLUX_COL.name())
             .setErrorBarType(ErrorBarType.capped_lines)
             .setMarkType(ShapeType.open_circle)
             .setSize(4);
```
jbudynk: Why not keep this statement as setting the default plotter preferences, and move the `setxCol()` and `setyCol()` calls into `setInSource()` at line 189? That way, all the meat (x, y, and errors) is under `setInSource()`, and the plotter preferences (errorBarType, markType, size, etc.) are separate. Also, anytime `setInSource()` is called, the x and y values will be set as well.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/StilPlotterTest.java
```diff
@@ -98,7 +101,7 @@ public void testAddSed() throws Exception {
     public void testAddTwoSegments() throws Exception {
         
         sed = ExtSed.read(TestData.class.getResource("3c273.vot").openStream(), SedFormat.VOT);
-        sed.addSegment(ExtSed.read(TestData.class.getResource("test300k_VO.fits").openStream(), SedFormat.FITS).getSegment(0));
+        sed.addSegment(ExtSed.read(TestData.class.getResource("test.vot").openStream(), SedFormat.VOT).getSegment(0));
```
jbudynk: Why did this SED get changed? Is it because it took too long to load?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/SegmentLayer.java
eholum: That's a really good point - though maybe we should just get rid of the setters until we're ready to plot any column? 

This column stuff gave me a headache, but basically with iris star tables, we want every column name to be fixed in the envpreferences, so xcol and ycol in the environment preferences should always be set to the spectral and flux column values. The error columns should only be specified if there's data available - which is kind of lame, but that was the way that I came up with to do it. Though since the names are fixed, maybe it's not even worth having setters for those values in the first place.

Ultimately when we actually put in logic for column selection that lets users plot whichever column they want, the set*col should actually call to the IrisStarTable to put the data in the appropriate column. But since that's not going to be there yet, I think I should just remove the methods and leave it as a todo? Does that make sense?

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/StilPlotterTest.java
eholum: Yeah, not a huge difference, but I added the other file to the test repository so I figured we might as well use it. Makes that unit test 3 seconds faster.

jbudynk:
Do you think it makes sense to include a "resetUnits()" method in SedPreferences.java? It would reset the spectral and flux units back to each segment's original units. Something like

```
void resetUnits() {
    for (int i=0; i < sed.getNumberOfSegments(); i++) {
        Segment seg = sed.getSegment[i];
        MapKey me = new MapKey(seg);
        setUnits(seg, segmentLayers.get(me));
    }
}
```

I think it'd end up being public, since I imagine this being used in the Metadata Browser to reset the units.

Or, instead of introducing a resetUnits() method, we'd have columns containing the original flux, spectral values, and errors.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/SegmentLayer.java
jbudynk: Neat, sounds good

eholum:
It probably would be eventually - there's still quite a bit to manage around handling units in the plotter, and I didn't want to get too far down the road on this branch. For now, the original data values are still there, both in the segment and in the deserialized star table - with they're original column names. Though the metadata browser's scroll pane needs to be fixed to handle that.

eholum:
I updated the pr with a few additional comments, as well as a fix for https://github.com/ChandraCXC/iris/issues/258.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/SegmentLayer.java
jbudynk: Ah, I see. When you say "I should just remove the methods and leave it as a todo," do you mean specifically setting the X, Y, and error column names somewhere and removing all the column setter methods? If so, I think that's fine to do, given that the setter methods will eventually live in IrisStarTable. 

But honestly, I'm not too worried if we have the setters in the code or not for now. If you do decide to remove them, let's just make sure that we remember to add the setters at a later point. Maybe it's smart to put the TODOs in both the SegmentLayer and IrisStarTable classes (knowing me, I'd forget otherwise :) )

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/StilPlotterTest.java
jbudynk: Ok, sounds good.

jbudynk:
I added a few more lines to your fix in #258 so that the suffix numbering increments more naturally. Before it was incrementing like "m32, m32 1, m32 1 2, m32 1 2 3". The latest change should increment the suffixes like "m32, m32 1, m32 2, m32 3". It also adds a test that checks the segment layer names.

eholum:
Merged into sprint 5

Issue #258, opened 2016-02-12
Issue with plotting segments with the same Target name


jbudynk:
If two SED segments share the same Target name, only one of them is shown on the Plotter. I believe the top-most segment of the same Target name is shown. Both segments appear in the Metadata Browser.

Test data: [Iris/github-issue-attachments/258](https://drive.google.com/open?id=0BwbX0VxOVx-iT3hxYm1zbXNoY1E)

This is on the master branch at 646227a, but is also present in sprint-5 at 8fb56eb

eholum:
Closed in https://github.com/ChandraCXC/iris/pull/257

PR #260, opened 2016-02-16
Iris dev 14 implement automatic color selection

jbudynk:
Addresses ChandraCXC/iris-dev#14

This PR introduces a very simple automatic coloring for added Segments. Using an HSV cylinder as the color space, the golden ratio is used to create a set of fairly distinct colors. I say "fairly" because only the hue changes, not the brightness or saturation. The colors have full brightness and saturation. In commit e2c6522, I tried to change the brightness and saturation every six colors or so to try to get more variety. However, this algorithm produced a lot of pale yellow colors, which is not wanted.

I created a ColorPalette interface to use for other color palettes we may create later.

Issue ChandraCXC/iris#258 causes problems for Segment coloration. If two or more segments have the same Target Name, the SegmentLayer colors are replaced with the latest ones, even if you remove the newly added segments from the plot.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/HSVColorPalette.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/HSVColorPalette.java
@@ -56,17 +56,6 @@ public class HSVColorPalette implements ColorPalette {
         return colors;
     }
     
-//    public Color getNextColor() {
-//        
-//        // make sure no color is too close to white, yellow, or black
-//        final float hue = random.nextFloat();
-//        // Saturation between 0.1 and 0.3
-//        final float saturation = (random.nextInt(2000) + 1000) / 10000f;
-//        final float luminance = 0.9f;
-//        
-//        return Color.getHSBColor(hue, saturation, luminance);
-//    }
-    
     /**
      * Returns a new distinct color using the golden rule ratio.
      * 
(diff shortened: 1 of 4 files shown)


iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/HSVColorPalette.java
```diff
@@ -0,0 +1,115 @@
...
+//            saturation -= -0.25;
+//        }
+//        
+//        // counter for hues.
+//        index++;
+        
+        return color;
+    }
+    
+    public final static String toHexString(Color color) throws NullPointerException {
```
eholum: What about making the color palette abstrac and putting static methods there?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/HSVColorPalette.java
```diff
@@ -0,0 +1,115 @@
...
+package cfa.vo.iris.visualizer.plotter;
+
+import java.awt.Color;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+
+/**
+ *
+ * Generate N distinct colors
```
eholum: Can you add a wiki link or something to the algorithm description?

eholum:
Sweet! Looks good!

I'm not sure about that issue with only one star table showing up. It's possible that plot2plane will only plot one startable with a given name...

jbudynk:
Thanks for the feedback! I've folded them in.

eholum:
Merged into sprint-5

Issue #13, opened 2014-05-19
Spectral data from FITS binary table won't display - Iris [BR-161]


jbudynk:
Bug report from ESO.

Loaded VO-like FITS binary tables of a spectra with ~9000 data points into Iris 2.0. It took 10-20 minutes to load the spectra, which we expect, since Iris 2.0 cannot handle spectra efficiently. However, he was unable to display the data in the SED Viewer, which is not expected.

One of his files FITS files is attached, along with a screenshot.

Here's his original email:

I need to report a problem using IRIS.

We tried to load and display in IRIS (v2) a VO-like FITS binary table 1d spectrum, like the one attached. The problem is that IRIS took anything between 10 and 20 minutes to just load the spectrum into memory (using all the cpu available), and I never succeeded displaying it (the only thing I get is the IRIS VAO SED ANALYSIS TOOL logo window (see attached screen shot)).

Both Linux (3.3.7-1.fc17.x86_64, java 1.6.0_22) and Mac (10.7.5, java 1.7.0_13-b20) gave identical results.

It would be very nice if you could have a look.

FITS file used: https://drive.google.com/file/d/0BwbX0VxOVx-iNWc1bEJsVkJvYnM/edit?usp=sharing

PR #261, opened 2016-02-22
iris-dev #13  Global Plot Preferences Management

jbudynk:

This PR adds global plot preference management to Iris. The end result is intended to include:

 1. a way to edit the plot preferences from the Plotter menus
 2. a plot preferences widget
 3. the widget integrated into the Plotter via a preferences frame

Currently, (1) is almost finished. This is the baseline necessary for meeting the Tier 1 stories from the BRD. As of commit 5c87238, the following preferences can be specified:

 - Plot type (log, linear, xlog, ylog)
 - Grid lines on/off

I need to do:

 - show/hide error bars
 - set auto/fixed plot range

(2) and (3) are Tier 2 tasks as Iris 2.1 does not have these tasks. However, I've started on (2), and (3) depends on ChandraCXC/iris-dev#21.


jbudynk:
closing this because it's based on the wrong branch

PR #262, opened 2016-02-22
iris-dev #13 - Global Plot Preferences Management

jbudynk:

This PR adds global plot preference management to Iris. The end result is intended to include:

 1. a way to edit the plot preferences from the Plotter menus
 2. a plot preferences widget
 3. the widget integrated into the Plotter via a preferences frame

Currently, (1) is almost finished. This is the baseline necessary for meeting the Tier 1 stories from the BRD. As of commit 5c87238, the following preferences can be specified:

 - Plot type (log, linear, xlog, ylog)
 - Grid lines on/off

I need to do:

 - show/hide error bars
 - set auto/fixed plot range

(2) and (3) are Tier 2 tasks as Iris 2.1 does not have these tasks. However, I've started on (2), and (3) depends on ChandraCXC/iris-dev#21.


jbudynk:
The behavior of these preferences is that they are truly "global." Meaning, if you change one SED plot view to xlog, then create a new SED and show it in the Plotter, that new SED will also be plotted in xlog space. 

This is not what I want to happen, though I'm open for a discussion. I think these global preferences should still be SED-specific, but apply to the *whole* SED, and not any specific Segments.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPreferences.java
```diff
@@ -30,6 +30,27 @@
...
+    //public static final String SHOW_ERRORS = "show_errors"; // not STILTS
+    
+    // Plot Types - Iris-specific, not STILTS.
+    public enum PlotType {
+        LOG("log"),
+        LINEAR("linear"),
+        X_LOG("xlog"),
+        Y_LOG("ylog");
+        
+        public String name;
```
eholum: If we did keep this enum here, you could add boolean values for x and y log, then below when you're setting them you wouldn't need the switch. Just be setXLog(plotType.xlog); setYLog(plotType.ylog);

jbudynk:
Latest changes:

 - PlotPreferences are created for each SED in SedPreferences. When a SedEvent occurs, the PlotPreferences object in VisualizerComponentPreferences updates to the selected SED. Now setting one SED to linear space, and another to log space doesn't affect the other.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPreferences.java
jbudynk: Check commit 528de73. Is this what you had in mind? Certainly makes the code shorter.

jbudynk:
I rebased this branch off of master at commit 0cee6a4.

I also migrated the PlotterView to Netbeans. The changes that occurred between 528de73 and d597064:

 - StilPlotter is a bean now. This was so I could create the PlotterView with Netbeans. It has a 0-argument constructor. 
 - The VisualizerComponentPreferences property in StilPlotter is no longer final.
 - Initializing the StilPlotter in PlotterView has changed. It is a little messy; I plan to refactor it to make it simpler.

TODOs:
 - add plot preference bindings to PlotterView components.
 - add tests to make sure the plot preferences update as expected.
 - simplify StilPlotter initialization in PlotterView.

eholum:
Looks good! Thanks for getting the plotter moved over to netbeans, I can open it in 8.0.2 just fine. I think there's some good work we could do moving more model definitions out of the metadata browser and plotter and into some set of shared model classes, but we can talk about that as we implement more cross-functionality between the two. This looks good to merge! But not into the sprint-6 branch...

jbudynk:
oh, haha, yes I'll open a new PR to merge into sprint-7 when I'm ready to merge.

jbudynk:
Updates:

Most of this work was for implementing a fixed plot view, meaning if the user selects "Fixed" from the View menu, the plot will remain at the latest viewpoint when plot changes occur. For example, if you zoom in on a plot, set the plot to Fixed, and add a new SED, the plot ranges will remain the same; the default behavior would update the plot window to the full data range. The viewport will also remain when you switch between SEDs.

One caveat to this is you can't switch Plot Types at the moment. Adding this capability in is on my TODO list.

I also added zoom in/out functionality to the zoom in/out buttons on the PlotterView, but these are not working properly... if anyone has a good reference to a zoom in/out algorithm, I'm all ears.

Grid line properties have also been bound to the selected SEDs.

TODOS:
----------
 - handle fixed-plot views when switching plot types (linear, log, xlog, ylog)
 - fix zoom in/out button algorithm
 - close this PR and open a new one for sprint-7

jbudynk:
Latest update allows plot type changes (between linear, log, etc.) while in fixed-plot mode, and plot types update to the SED's plot preference when switching to the SED.

jbudynk:
I'm closing this PR and re opening it for sprint-7. It is now in PR #270

jbudynk:
closed

PR #263, opened 2016-02-23
Split Reset Method #7

eholum:
Closes https://github.com/ChandraCXC/iris-dev/issues/7

Adds a segment event listener to the VisualizerComponentPrefereces settings to give the plotter/metadata browser more detailed information into sedmanager changes, allowing more efficient updates of the model and GUI. This change drastically improves performance in the plotter - at least while testing on my box and on my laptop.

This overlaps slightly with https://github.com/ChandraCXC/iris/pull/262, and this should probably be incorporated in some manner so that changing plot preferences doesn't reset the existing plot bounds.

Important note - we have access to the plotter bounds though the plane aspect object in the PlotDisplay.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
@@ -26,10 +26,14 @@ import cfa.vo.iris.IWorkspace;
 import cfa.vo.iris.events.SedCommand;
 import cfa.vo.iris.events.SedEvent;
 import cfa.vo.iris.events.SedListener;
+import cfa.vo.iris.events.SegmentEvent;
+import cfa.vo.iris.events.SegmentEvent.SegmentPayload;
+import cfa.vo.iris.events.SegmentListener;
 import cfa.vo.iris.sed.ExtSed;
 import cfa.vo.iris.visualizer.plotter.PlotPreferences;
 import cfa.vo.iris.visualizer.plotter.SegmentLayer;
 import cfa.vo.iris.visualizer.stil.IrisStarTableAdapter;
+import cfa.vo.sedlib.Segment;
 
 /**
  * Single object location for data and preferences needed by the iris visualizer 
@@ -59,10 +63,11 @@ public class VisualizerComponentPreferences {
         this.plotPreferences = PlotPreferences.getDefaultPlotPreferences();
         
         // Add SED listener
-        addSedListener();
+        addSedListeners();
     }
     
-    protected void addSedListener() {
+    protected void addSedListeners() {
+        SegmentEvent.getInstance().add(new VisualizerSegmentListener());
         SedEvent.getInstance().add(new VisualizerSedListener());
     }
     
@@ -120,7 +125,7 @@ public class VisualizerComponentPreferences {
     }
 
     /**
-     * Adds the SED to the preferences map.
+     * Adds or updates the SED to the preferences map.
      * @param sed
      */
     public void update(ExtSed sed) {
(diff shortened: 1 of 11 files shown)


iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
```diff
@@ -120,14 +124,28 @@ public SedPreferences getSedPreferences(ExtSed sed) {
...
+     * Adds or updates the segment within the specified SED.
+     * @param sed
+     * @param segment
+     */
+    public void update(ExtSed sed, Segment segment) {
+        if (sedPreferences.containsKey(sed)) {
+            sedPreferences.get(sed).addSegment(segment);
         } else {
-            sedPreferences.get(sed).refresh();
+            sedPreferences.put(sed, new SedPreferences(sed, adapter));
```
jbudynk: Is there anything to be done with the Segment if the SED isn't in the map? Should it also be added to the SED before it gets added to the sedPreferences map?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
```diff
@@ -137,8 +155,24 @@ public void update(ExtSed sed) {
      * @param sed
      */
     public void remove(ExtSed sed) {
-        sedPreferences.remove(sed);
+        if (sedPreferences.containsKey(sed)) {
+            
+        }
```
jbudynk: This isn't doing anything. Did you mean to put the lines below inside this statement?

jbudynk:
Nice work on getting the plotter to work faster! I can load in the 300k point file, and I no longer see any lag while moving the plotter around. (on Linux). I can also confirm that adding a new segment to the same SED doesn't take a long time, as the plotter no longer recreates the dataset every time a SED is changed.

I have a question about the REDRAW capability. Say I want to fix the plot ranges when adding an SED. Is this possible? Or is this more for switching the grid lines, log/linear plot scales, things that don't affect the actual SED data?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
eholum: whoops, yup. supposed to be a return statement in there but it got dropped. Thanks!

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
eholum: That's a good point, it's pretty unclear.

The notifications in ExtSed for add/update imply that the segment is already apart of the ExtSed, then the SedPreferences constructor will handle creating the segment layer since the segment is already there. I'll take a look and either adjust this code to make it more clear, or just add a comment.

eholum:
Yeah that's definitely possible. And should already be supported, just need to modify the change listeners in the visualizer to use the correct methods.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
jbudynk: I think a comment would suffice

jbudynk:
Nice. I'm thinking that this will be useful in implementing a fixed viewport even when SEDs change for ChandraCXC/iris-dev#13.

eholum:
Cleaned up the things mentioned above, as well as the reset methods in the stil plotter. More work to be done there in terms of functionality but this accomplishes the goals laid out in the task.

jbudynk:
Looks ok to merge to me. Just a few questions:

 1. Why is the data no longer being cached?
 2. Have you noticed any differences in performance between c38a53f and 4a6b54d due to caching? I can't seem to see a difference, though I thought on Wednesday I noticed a significant difference. But now I'm testing it out and I can't see any differences... so perhaps this is a null issue :)

eholum:
I don't think I changed the caching behaviour? Still setting everything with https://github.com/ChandraCXC/iris/pull/263/files#diff-62c68282ac1c34f74a07fa31fb632756R133

And calling reset with true. Ultimately I don't think it matters either way for us since we aren't really using repaints. In the future that might change, but for now might as well leave as is since I didn't see any performance issues, either.

jbudynk:
Oh you know what? The caching changed only between  c38a53f and 4a6b54d. But it looks like it's back to what it was before this PR, which is not caching the data. See https://github.com/ChandraCXC/iris/pull/263/files#diff-d16d4d6a7d7ec5844bfa47dedeac896eL188 between those two commits. I think 4a6b54d had `reset(sed, false)` while c38a53f brings it back to `reset(sed, true)`. So a better question would've been why did you change it at 4a6b54d, then change it back in c38a53f?

But as you say,  it probably doesn't matter anyhow haha. So, I'm happy with a merge into sprint-6 :)

PR #264, opened 2016-03-01
Using Tabs in the Metadata Browser

eholum:
For
https://github.com/ChandraCXC/iris-dev/issues/33
https://github.com/ChandraCXC/iris-dev/issues/30
https://github.com/ChandraCXC/iris-dev/issues/27

I went through and remade the metadata browser in netbeans, as windowbuilder in eclipse doesn't generate the form file that netbeans requires. There was enough changing in the UI that I don't think it was that big of a deal, and now we all should have access to the netbeans editor for binding, etc.

This PR splits the MB into 3 panels, one for the plotted data, one for all the available data, and one for all the segment metadata. It also handles getting the column sizes in each tab big enough to read the names of the columns without manually having to expand them. It also uses binding to handle the MB gui wherever possible. Though that may get more complicated (hopefuly not) in the future.

There are two large items remaining in terms of the UI - first is being able to view multiple segments at a time (right now it's single selection only), the other is viewing segment metadata, for which I have not yet come up with a method for getting table parameters out of a star table, but I'm hoping there's something already built in stil for it. The former will be taken care of in https://github.com/ChandraCXC/iris-dev/issues/20, along with some refactoring of the list selection methods.

For now let me know if there are any high level issues with the GUI, or if there are any performance issues that you may see with the new interface.


jbudynk:
Looks good! Glad to see that you can scroll left and right in the browser. 

I don't see any performance issues with the MB once the data is loaded in. As a test, I loaded the 300k-point file into Iris. Expectedly, this took quite some time (on the order of 15 s on Linux, and 25 on OS X (!!)), but once it was loaded, everything worked like a charm. I could quickly scroll through the 300k data points in the browser, and switch tabs.

Another thing I like is that if I close the window and reopen the MB, highlighted points are persistent. If I switch to another SED and back, the highlighted points are gone, but I can live with that. I wouldn't have expected the points to remain highlighted after switching SEDs.

I don't know if I like the name "Plotter Data" for the data tab, as it sounds like only the data shown in the plotter window shows up here (like if you're zoomed-in on a region, only those data points would show in the MB). This tab is supposed to show only the "meat" of the data: flux, spectral value, errors, and original flux, as it currently does. The behavior is right, but the tab name might be misleading.

Ah! which brings up another missing element. The original flux values should also be included. This should be another column in the Plotter Data tab. This may be better addressed in a new task if we want to discuss how to keep track of original fluxes, but we should have a column for the original flux.

Are the indices between the "Plotter Data" and "Point Metadata" tabs always the same? I ask because down the road, we'll want to add filters. In Iris 2.1, after applying a filter in either the data or metadata tab would select the same point in the other tab.

jbudynk:
Also, the metadata browser might be a little too tall for the default Iris desktop window. On OS X and Linux, I notice the bottom of the MB falls off the bottom of the desktop.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/MetadataBrowserMainView.java
```diff
@@ -0,0 +1,484 @@
...
+        setChangeListener();
+
+        resetData();
+    }
+
+    protected void setChangeListener() {
+        VisualizerChangeEvent.getInstance().add(new MetadataChangeListener());
+    }
+
+    public synchronized void resetData() {
```
jbudynk: [minor note] -- from the old MetadataBrowserView.java `reset()` method, you had a note saying why this method should be synchronized:

```
// Multiple threads resetting this simultaneously can do weird things
// to JList and JTable objects.
```

should the note remain here in case down the road someone wonders why it's synchronized?

eholum: Hmm. Does that method even need to be synchronized anymore?

eholum:
+ I shrank the size of the default window so that it fits on my mac.
+ Added a very rudimentary method for viewing segment metadata. It's crude, but at least it lets us see what's in there. 

In response to above:

Any suggestions on a better name for the "plotter data" tab? Should it just be Data as it was in Iris 2.1?

In regards to the 'original values'. There's still some work to do with what's in the Segment Star Tables, including adding original value columns. I didn't include that work here since we'll be able to amend those star tables at a later time. There's some work to do in the iris data model, so holding off on that till that is done. On the plus side, once that's done - since it's all star tables we don't have to worry about changing anything here.

As of now they are since the data and plotter tables are guaranteed to have their rows line up. Though the filter work is going to have to check that the indicies are correct. For now I just dumped them in.

jbudynk:
> Any suggestions on a better name for the "plotter data" tab? Should it just be Data as it was in Iris 2.1?

I'm good with either "Data" or "Point Data." @olaurino, any thoughts?

> In regards to the 'original values'. There's still some work to do with what's in the Segment Star Tables, including adding original value columns. I didn't include that work here since we'll be able to amend those star tables at a later time. There's some work to do in the iris data model, so holding off on that till that is done. On the plus side, once that's done - since it's all star tables we don't have to worry about changing anything here.

Agreed. Definitely makes sense to work this in another task. [Is there a task already open for this?]

> As of now they are since the data and plotter tables are guaranteed to have their rows line up. Though the filter work is going to have to check that the indicies are correct. For now I just dumped them in.

Sweet. Sounds good.

eholum:
Fixed some of the issues here.

Also I added work for https://github.com/ChandraCXC/iris-dev/issues/33, it's not a lot of code and it's pretty dependent on the PR, so throwing it in here as the last 2 commits.

Basically just added support to make the Select All and Invert Selection buttons to work. Additionally selecting points in the data tab selects the same points in the metadata tab - and vice versa.

jbudynk:
The invert selection works as expected for the Data and Point Metadata tabs. However, it shouldn't be highlighting anything in the Segment Metadata tab. The Segment Metadata tab should be independent of the selections in the other tabs. For example, if I select a row in the Segment tab, then click "Invert Selection," and switch to the other tabs, you'll see all of the rows are selected. Instead, the other data tabs should remain as they were before touching the Segment tab.

at some point, we'll want to add an "Unselect All" button. For now, you can just click "Select All" and "Invert Selection" to clear the highlights.

Otherwise, looks good!

eholum:
Yeah, that makes a lot of sense. I removed that logic that selected elements in the segment tabs. I also added a clear selection button.

eholum:
Sorry to keep expanding this PR, wanted to address the comment above about how long it takes to load in the 300k fits file. In particular, it was incredibly bothersome that the UI because unresponsive for 20 seconds - so I made the VO serialization asynchronous. The plotter data is serialized first, and it happens very quickly, so the UI no freezes for only a second or two while the plotter is loaded. When the executor finishes its task it updates the placeholder and notifies the metadata browser to reload it's data.

eholum:
Talked with Jamie during standup - and have fixed the button selection logic in the segment panel.

PR #266, opened 2016-03-09
iris-dev#19 Viewing Multiple Segments in the Metadata Browser

eholum:
This PR closes

https://github.com/ChandraCXC/iris-dev/issues/19
https://github.com/ChandraCXC/iris-dev/issues/20

And additional deals with making the metadata browser/visualizer component more stable in general.

We include a new class, StackedStarTable, which is essentially a http://www.star.bristol.ac.uk/~mbt/stil/javadocs/uk/ac/starlink/table/ConcatStarTable.html that also maps all the columns in the provided list of startables into a single metadata star table. Users provide a column matcher class which determines whether or not two columns should "line up". In our case we have implemented a single class, the UtypeColumnInfoMatcher, which may or may not need to be improved upon in the future. For now it appears to line data up well in the test cases that I have provided.

This is achieved through wrapping StarTables into several different other star tables provided by both stil and iris. In particular, a ColumnMappingStarTable uses a ColumnPermutedStarTable and ConstantColumns/JoinStarTables to map an existing StarTable's columns into a provided metadata star table.

e.g.

Provided a metadata table with columns, A - B - C - D, and a data table with columns D - B, the ColumnMappingStarTable maps the baseTable columns to the metadata columns, and turns the dataTable to something looking like, * - B - * - D, where the * are empty columns. From there, it is a simple matter to get a PrimaryMetadataTable by simply creating a set by way of a ColumnMetadataStarTable - which takes a list of startables and produces the smallest columninfo set containing all the tables' columns. Important to note that all of this is generic to StarTables, rather than Iris.

The metadatabrowser now uses StackedStarTables in place of regular StarTables in it's data panels to support viewing of multiple segment's data simultaneously.

I also added some work in refactoring IrisStarTables/SegmentStarTables to make them easier to read and more robust to errors. SegmentStarTables are now the primary data holders for all information related to plotting - and can be extended to be more dynamic than they are now when we are actually interested in working with double[] of data. This was done in an effort to make filtering tasks easier, but there is still more work to be done there.

Lastly was some minor clean up to error handling and the addition of a few more unit tests to verify these changes.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
+++ b/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
@@ -54,6 +54,7 @@ public class SegmentStarTable extends ColumnStarTable {
 
     private double[] specValues;
     private double[] fluxValues;
+    private double[] originalFluxValues;
 
     private double[] specErrValues;
     private double[] specErrValuesLo;
@@ -84,6 +85,7 @@ public class SegmentStarTable extends ColumnStarTable {
         
         setSpecValues(segment.getSpectralAxisValues());
         setFluxValues(segment.getFluxAxisValues());
+        setOriginalFluxValues(segment.getFluxAxisValues());
         
         // Try to add flux error values
         this.fluxErrValues = (double[]) getDataFromSegment(Utypes.SEG_DATA_FLUXAXIS_ACC_STATERR);
@@ -138,6 +140,9 @@ public class SegmentStarTable extends ColumnStarTable {
         if (fluxValues != null) {
             setFluxValues(units.convertY(fluxValues, specValues, fluxUnits, specUnits, newUnit));
         }
+        if (originalFluxValues != null) {
+            setOriginalFluxValues(units.convertY(originalFluxValues, specValues, fluxUnits, specUnits, newUnit));
+        }
         if (fluxErrValues != null) {
             setFluxErrValues(units.convertY(fluxErrValues, specValues, fluxUnits, specUnits, newUnit));
         }
@@ -166,19 +171,22 @@ public class SegmentStarTable extends ColumnStarTable {
     void setSpecValues(double[] specValues) {
         removeColumn(Column.SPECTRAL_COL);
         this.specValues = specValues;
-        
-        if (!isEmpty(specValues))
-            addColumn(PrimitiveArrayColumn.makePrimitiveColumn(
-                Column.SPECTRAL_COL.getColumnInfo(), (Object) specValues));
+        addColumn(PrimitiveArrayColumn.makePrimitiveColumn(
+            Column.SPECTRAL_COL.getColumnInfo(), (Object) specValues));
     }
 
     void setFluxValues(double[] fluxValues) {
         removeColumn(Column.FLUX_COL);
         this.fluxValues = fluxValues;
-        
-        if (!isEmpty(fluxValues))
-            addColumn(PrimitiveArrayColumn.makePrimitiveColumn(
-                Column.FLUX_COL.getColumnInfo(), (Object) fluxValues));
+        addColumn(PrimitiveArrayColumn.makePrimitiveColumn(
+            Column.FLUX_COL.getColumnInfo(), (Object) fluxValues));
+    }
+    
+    void setOriginalFluxValues(double[] originalFluxValues) {
+        removeColumn(Column.ORIGINAL_FLUX_COL);
+        this.originalFluxValues = originalFluxValues;
+        addColumn(PrimitiveArrayColumn.makePrimitiveColumn(
+            Column.ORIGINAL_FLUX_COL.getColumnInfo(), (Object) originalFluxValues));
     }
 
     void setSpecErrValues(double[] specErrValues) {
@@ -217,12 +225,14 @@ public class SegmentStarTable extends ColumnStarTable {
                 Column.FLUX_ERR_LO.getColumnInfo(), (Object) fluxErrValuesLo));
     }
     
-    private void removeColumn(Column column) {
+    private int removeColumn(Column column) {
         for (int i=0; i<columns.size(); i++) {
             if (column.name().equals(getColumnData(i).getColumnInfo().getName())) {
                 columns.remove(i);
+                return i;
             }
         }
+        return -1;
     }
 
     /*
@@ -282,7 +292,8 @@ public class SegmentStarTable extends ColumnStarTable {
         SPECTRAL_ERR_LO("X axis low error values", "iris.spec.value"),
         FLUX_COL("Y axis values", "iris.flux.value"),
         FLUX_ERR_HI("Y axis error values", "iris.flux.value"),
-        FLUX_ERR_LO("Y axis low error values", "iris.flux.value");
+        FLUX_ERR_LO("Y axis low error values", "iris.flux.value"),
+        ORIGINAL_FLUX_COL("Original flux values", "iris.flux.value.original");
         
         public String description;
         public String utype;
--- a/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
+++ b/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
@@ -16,14 +16,16 @@
 
 package cfa.vo.iris.sed.stil;
 
+import java.io.IOException;
 import java.util.UUID;
 import java.util.logging.Logger;
 
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.StringUtils;
+
+import uk.ac.starlink.table.ColumnData;
 import uk.ac.starlink.table.ColumnInfo;
 import uk.ac.starlink.table.ColumnStarTable;
-import uk.ac.starlink.table.ConstantColumn;
 import uk.ac.starlink.table.PrimitiveArrayColumn;
 import cfa.vo.iris.units.UnitsException;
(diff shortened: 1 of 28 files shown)


iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
```diff
@@ -71,191 +82,276 @@ public SegmentStarTable(Segment segment, String id)
...
+    
+    public ColumnData getColumnData(int index) {
+        if (index == 0) {
+            return nameColumn;
         }
-        return ret;
+        
+        Iterator<SegmentDataColumn> it = columns.iterator();
+        for (int i=0; i<index-1; i++) it.next();
+        return it.next();
```
jbudynk: what happens if the index is outside the range of columns?

eholum: Ah, excellent points. i'll add a check and exception of the index is greater than the number of columns. There's no check in a column star table either (https://github.com/Starlink/starjava/blob/83a1eef2bb85ff87772b688803cec44f98745665/table/src/main/uk/ac/starlink/table/ColumnStarTable.java#L94), so I'll change the code that uses those to have a check as well.

iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
```diff
@@ -71,191 +82,276 @@ public SegmentStarTable(Segment segment, String id)
...
+     * valued columns in the table.
+     * 
+     * @param newUnit
+     * @throws UnitsException
+     */
     public void setFluxUnits(YUnit newUnit) throws UnitsException {
         if (specValues == null) return;
         
+        // Convert units
         if (fluxValues != null) {
```
jbudynk: why are the flux-related variables checked for 'not null', but not the spectral-related variables?

iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
```diff
@@ -71,191 +82,276 @@ public SegmentStarTable(Segment segment, String id)
...
-        
-        if (!isEmpty(specValues))
-            addColumn(PrimitiveArrayColumn.makePrimitiveColumn(
-                Column.SPECTRAL_COL.getColumnInfo(), (Object) specValues));
+        updateColumnValues(specValues, Column.Spectral_Value);
     }
 
-    void setFluxValues(double[] fluxValues) {
-        removeColumn(Column.FLUX_COL);
+    public double[] getFluxValues() {
```
jbudynk: the rest of the getters/setters are public, unlike the spectral-related values (and the comment near line 234). is this on purpose?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/MetadataBrowserMainView.java
```diff
@@ -108,61 +118,109 @@ private void setDataTables() {
...
+    }
+    
+    
+    private void updateDataTables() {
+        List<StarTable> plotterDataTables = new LinkedList<>();
+        List<StarTable> segmentDataTables = new LinkedList<>();
+        
+        for (IrisStarTable table : selectedStarTables) {
+            plotterDataTables.add(table.getPlotterTable());
+            segmentDataTables.add(table.getDataTable());
```
jbudynk: I wonder if it makes sense to rename the IrisStarTable get/setDataTable() methods to get/setSegmentDataTable(), given the naming conventions used in this file? In other words, when I see `table.getDataTable()`, I don't think "Segment metadata," I think "spectral, flux, and error values"

iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
eholum: I think because of a prior implementation of units managment that would have thrown a null check if the y values were null. That doesn't seem to be the case anymore - I removed the checks.

iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
eholum: Time conflict between this PR and the next one. They should all be public, and I adjusted the comment.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/MetadataBrowserMainView.java
eholum: Changed

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/tables/VisualizerStarTableTest.java
```diff
@@ -0,0 +1,41 @@
+package cfa.vo.iris.visualizer.stil.tables;
+
+import org.junit.Before;
+
+import cfa.vo.sedlib.common.Utypes;
+import uk.ac.starlink.table.ColumnInfo;
+
+public abstract class VisualizerStarTableTest {
```
jbudynk: there are no tests run here. It'd be interesting to see a testcase for UTypes that have a different prefix (e.g., handling the case of `spec:SpectralAxis.foo.bar` vs `phot:SpectralAxis.foo.bar`).

eholum: This is just an abstract test case for sharing/setting up common data between some of the other tests in the package.

I'll add a test for it, we have a task in our backlog for finding additional test files that cover the ranges of issues we're likely to see. Also SedLib doesn't include equality checks for prefixes that aren't "spec", unfortunately, I had something in place before to override that - and I can put that back in so that those cases line up.

eholum: Also there wasn't a task for the invalid files list, but I added it. https://github.com/ChandraCXC/iris-dev/issues/50

jbudynk:
I've finished reviewing the PR. Looks good for the most part! Definitely more clever than Iris' current metadata browser :) 

Most of my comments you've already addressed (thanks)

I did some testing just fooling around with datasets in the MB. [These are the datasets I used](https://drive.google.com/open?id=0BwbX0VxOVx-iRG1GVmVmTnpGa1U) (Google Drive link).

**RESULTS**
The performance is good. I don't notice any lag in the MB.


**BUGS**
One bug I noticed:  If I make the MB smaller in width, I can't scroll left-and-right in the Data tabs. The same thing happens in the Point Metadata tab, which is weird since I see a scroll bar, but I can't scroll all the way to the right side, and the vertical scroll bar disappears.

**CONCERNS**
I have a few concerns, though I'm not sure they belong to this PR or a previous one. 

1. The column names of the Point Metadata tab change to whatever the actual column names are of the *top-most selected segment*. I would think the x, y, and yerr column names should be either the IVOA standard column names (e.g., the UTypes), or like the names that are in the Data tab (Spectral_Value, Flux_Value, Flux_Error_Value). For example, if you load in the test data set, switch to the Point Metadata tab, and select the J113007.5-144927 segments, you'll find the column names are not informative.

2. The stacked StarTables aren't sorted by spectral value. I don't think that was a requirement we had at the creation of issue iris-dev#19, so I don't know if this needs to be done in this PR. However, it would make great sense if the rows were sorted by spectral value. It could be added in for or after PR #269.

3. Again, perhaps not for this task, but sorting the rows by clicking on a column would be a great and useful addition. (for this, PR #269, or after)

jbudynk:
Hmm... I also noticed that something seems off with the Segment tab. With the dataset used above, I see that a number of columns aren't populated at all. Examples are:

[from datasets created in, exported from, and reloaded back into Iris]
 - Extent
 - Min
 - Max
 - StatError

[from NED data; also IVOA UTypes]
 - SpatialAxisCoverageBoundsExtent
 - SpatialAxisCoverageBoundsExtent
 - TimeAxisCoverageBoundsExtent
 - SpectralAxisCoverageBoundsExtent
 - SpectralAxisCoverageBoundsStart / Stop

[from ASDC data]
 - SpaceFrameEquinox

Shouldn't these be cut-off from the data?

I'd also like to note that I tested the MB updates when I remove segments from the SedBuilder, and when I switch between SEDs.

eholum:
Re: the resizing bug -
I have added a change to address that in https://github.com/ChandraCXC/iris/pull/269, there were additional changes there that I didn't want to have to deal with in the new layout twice.


Re: 1 -
I changed the behavior of both the point metadata tab and the segment metadata tab to use utypes as column names, when available. This makes those datasets (and most datasets) more clear, and gives us an easy way to remove the behavior if we don't like it at a later time.

Re: 2 and 3 - 
I'm adding tasks for them, as they are beyond the scope of this change:
https://github.com/ChandraCXC/iris-dev/issues/51
https://github.com/ChandraCXC/iris-dev/issues/52

Re: The empty columns in the segment metadata tab-
I added a small bit of logic to remove empty parameters from the segment serialization code. I don't know why sedlib/stil are adding them, but they were easy enough to take out. There are no long null columns in that data tab.

jbudynk:
I like the changes. We might want to revisit the names for the x, y, and yerr columns in the Point Metadata tab at some point (they are a little long), but it's mounds better than it was before, especially for those with column names ID_1, ID_2, etc.

I say merge it

PR #267, opened 2016-03-10
Fitting tool implementation

olaurino:
# Description

This PR addresses several tasks in the fitting tool epic:
 * Resolve chandracxc/iris-dev#16 - Model Browser with search function
 * Resolve chandracxc/iris-dev#41 - Implement a Fit Configuration Expression Builder
 * Resolve chandracxc/iris-dev#42 - GUI for Fit Results
 * Resolve chandracxc/iris-dev#43 - Confidence Intervals
 * Resolve chandracxc/iris-dev#44 - Read Sherpa models from xml file
 * Resolve chandracxc/iris-dev#70 - Add Vanilla chi-squared statistic to Fitting Tool
 * Unlabeled - Integrate Custom Models Manager

Unfortunately, I couldn't manage to keep everything separate, so this is turning to be a huge PR. However, I took a lot of extra time to refactor, and I plan to add javadocs ;)

As I went by I also fixed/refactored stuff that was annoying during development, especially some SAMP tests that would prevent mvn to verify the build.

Here are some details to guide through the code review:

### Read Sherpa Models from xml file
The xml is [here](https://github.com/ChandraCXC/iris/pull/267/files#diff-2792cb94a79133d20842c4bbd281c0d8). Nothing interesting. However, many models had a bogus Minimum (something like -1.xxxE38) that I had to fix. Not sure whether the old Iris was bypassing the issue. Also, I found that there are two powerlaw models in Sherpa (`powlaw1d` and `powerlaw`) so I included the missing `powlaw1d`.

The file is read by a [Model Factory](https://github.com/ChandraCXC/iris/pull/267/files#diff-8476de34b127d4d32da9bdae8e6d4dfd). Here is the relative [unit test](https://github.com/ChandraCXC/iris/pull/267/files#diff-f7f6dd293afd64b563e41e088f1cd5cb).

### Fit Configuration Expression Builder
The relevant backend class is [ModelExpressionVerifier](https://github.com/ChandraCXC/iris/pull/267/files#diff-8c83eb438b5fba246dc51c7b3cd00f4a), with its [unit test](https://github.com/ChandraCXC/iris/pull/267/files#diff-979805ed0a72896dd677bd9fc7f82d76)

The class is used to create an `InputVerifier` in [ModelViewerPanel](https://github.com/ChandraCXC/iris/pull/267/files#diff-feea0220291cfe19b39c5ca68c985dd6) as well as to provide validation to the [FitConfiguration](https://github.com/ChandraCXC/iris/pull/267/files#diff-78f0ee5e8c568d4f68442e6abd7f4ae6) class.

The validator uses the embedded JavaScript engine included in Java and binds model components to JavaScript numerical variables using their identifiers. If the engine returns a number after the binding and the evaluation of the model expression, then the model expression is valid.

While this provides a rather straightforward implementation, it would allow users to use JavaScripts'  Math library and do not get notified that the model is invalid. Sherpa, however, would the throw an exception that is gracefully presented to the user.

### Fitting GUI
A single View provides (most of, for now) the functionality of Iris 2.1.

The JInternalFrame containing the whole view is [FittingMainView](https://github.com/ChandraCXC/iris/pull/267/files#diff-ff330886aa3ee19e24520f363ae5408e). I tried to stick as much as possible to the MVC paradigm: the Model is basically the currently selected [ExtSed](https://github.com/ChandraCXC/iris/pull/267/files#diff-ff330886aa3ee19e24520f363ae5408e), that now has a [FitConfiguration](https://github.com/ChandraCXC/iris/pull/267/files#diff-78f0ee5e8c568d4f68442e6abd7f4ae6) property. I had to read and write the fit configuration enough times to realize soon that this was better than having to rely on the unsafe attachment mechanism.

In accordance to Iris 2.1 requirements, there is currently a 1:1 relationship between FitConfigurations and Seds. In the future, the relationship may be of the n:m kind.

The main fitting interface has a couple of unit test classes, [here](https://github.com/ChandraCXC/iris/pull/267/files#diff-a8ebfac53016d4d9bd74365516e34706) and [here](https://github.com/ChandraCXC/iris/pull/267/files#diff-bddade66543d4e6de897aa12c50d2227). A complex thread-like functional test that starts from setting up custom models and finishes computing the confidence limits is [here](https://github.com/ChandraCXC/iris/pull/267/files#diff-db2d6bb384ada5d31e590b6331efc4ab). Eventually, I would like to implement a drag and drop mechanism for "using" new models, but for now models can be used by double clicking on them.

The nested components of the main view are:
  * [ModelViewerPanel](https://github.com/ChandraCXC/iris/pull/267/files#diff-feea0220291cfe19b39c5ca68c985dd6). This is a common "widget" shared by the fitting tool and by the Science tool, although for the Fitting tool the panel is editable and represents the main panel for entering and visualizing the fit. It has a clumsy status panel that should grow into a better system for providing feedback to the user.
  * [ModelParameterPanel](https://github.com/ChandraCXC/iris/pull/267/files#diff-eb640c37e8c33cd31c8229565876de7b) is the actual component that displays the parameter values. It has a [unit test](https://github.com/ChandraCXC/iris/pull/267/files#diff-69cf243b35fb0d526cc5c7a438c61741)
  * [FitResultsPanel](https://github.com/ChandraCXC/iris/pull/267/files#diff-6d48be41b1174ab15434390aa34d3798) shows the statistics information after a fit.
  * [ConfidencePanel](https://github.com/ChandraCXC/iris/pull/267/files#diff-bf64f32e6e04ced08b9d60f4d5a23555) provides input and results visualization for the confidence limits. It has its own [unit test](https://github.com/ChandraCXC/iris/pull/267/files#diff-676e091c243a3d92189b6ce5e9e96e7d)
  * [FitController](https://github.com/ChandraCXC/iris/pull/267/files#diff-77bd0f698114655b966ae9ad8de33704) is the controller for the fitting capabilities. It does not have the usual features of a CRUD controller. It's more like a service implementation for performing fits. It delegates the logic for managing models to the [ModelsController](https://github.com/ChandraCXC/iris/pull/267/files#diff-d0b19b5ccc3cc5fc1b3941008751351d): while the list of preset models is static, the list of custom models can change at run time. The FitController listens for new custom models being installed or edited, combines them with the preset models, and updates the [ModelsTreeModel](https://github.com/ChandraCXC/iris/pull/267/files#diff-e03150aa54f688bc23f399e1d716ebe5) that backs the `JTree` of available model components in the View. Similarly, the [CompositeModelTreeModel](https://github.com/ChandraCXC/iris/pull/267/files#diff-1efd880cbb2b229dddc136848e216589) class backs the `JTree` that allows users to browse the model being built, changing the parameter values.

The custom models manager is pretty much the original component included in Iris 2.1. It was coded in an afternoon and integrated with Sherpa and the older visualizer in about 3 days. It is rather crude and should be improved in terms of usability and error handling.
  * [CustomModelsManager](https://github.com/ChandraCXC/iris/pull/267/files#diff-43f475e263eaae71038c1187e66e2e70)
  * [CustomModelsManagerView](https://github.com/ChandraCXC/iris/pull/267/files#diff-94408c1b5be1796683ecd9ce11905ffa)
  * [CustomModelsTreeModel](https://github.com/ChandraCXC/iris/pull/267/files#diff-7ca9988717c1701c8198c95d2177a3d0)

I seriously reviewed the [SherpaClient](https://github.com/ChandraCXC/iris/pull/267/files#diff-b49c2430797dae6d16393fb6c8fddb99), which is now a much more central piece of code, rather tha just backing up the smoke test. It has a simple [unit test](https://github.com/ChandraCXC/iris/pull/267/files#diff-6dddf2413f05b3bea50658dd23a735ea) to check that it produces a faithful `SherpaConfiguration` object from the `FitConfiguration`.

Note that the `samp-factory` backs the interaction between the Iris java GUI and Sherpa by means of a number of interfaces that capture the data model of the private communication interface over SAMP. This is the reason why there is a zoo of interfaces that are implemented by a dynamic proxy rather than actual classes: the dynamic proxy takes a bean-like instance, transaltes it into a Map that is sent over SAMP resulting in a result Map being sent back. The SAMP messages in the Iris/sherpa interface are particularly complex, which is not they way SAMP messages were intended to be (we even pass arrays serialized as base64 strings).The dynamic proxy provides a means to represent the SAMP messages by hiding the complexity of the nested maps.

In a few cases, where it made sense to provide an actual implementation, I did it, but I made sure to make clients aware of the interface only. There are some awkward constructs (e.g. a couple of static methods in [DefaultModel](https://github.com/ChandraCXC/iris/pull/267/files#diff-70027beb6ff8f6c69d85d5fa7986800c) that are required because the semantics of the SAMP messages is bizarre: e.g. `CompositeModel.name` is not a name but the model expression, and models have both a name and an id but they are both expressed in `Model.name` and `UserModel.name`. In other cases information is spread across different fields (e.g. paramer instances are split into array fields of a single instance) or even separate classes (e.g. a custom model is present both as a regular `Model` *part* of a `CompositeModel` and as a `UserModel` in a list that is represented by a different field in the `SherpaConfiguration`.

From the user's point of view, the most confusing part is that custom models must have a name that does not represents the name input by the user. In the end, we probably need specific adapters to make sure that the representation is user friendly while the messages corresponds to what Sherpa expects. In any case this limitation is already present in Iris 2.1. 

Sherpa client always fits synchronously. It is a task of the GUI to run the fits in a non-blocking way. Which brings me to the list of "known issues" of this PR:
  * The fit and confidence limits computation is performed synchronously in the main Swing thread. This si the the first thing to fix.
  * The sigma value in the confidence panel does not update.
  * Introduce layouts to stabilize the GUI and fix the size of the fitting view on OSX.
  * Improve feedback provided to the user (note that the GUI provided by this PR is already an improvement over Iris 2.1):
    * Messages like "Fit Failed" and "Invalid Model Expression" should be more visible.
    * When using Chi2, data without errors or with NaN as errors is being dropped. The only visual feedback is in the number of points visualized in the fit results.
    * Reset all panels when Seds are switched. This only happens under some circumstances.
  * The tree representing the model being fit should always be expanded. I did this for one of the trees but forgot to do it for the other.
  * Model expression could contain `Math.*` javascript operations and the model expression would be validated, but the fit would fail (again, Iris 2.1 does not provide expression validation, so this is an improvement already). On the other hand, the operation "to the power of" would invalidate the model but would still work with Sherpa.
  * The selection of models from the available models tree only updates the model description if one clicks on the model name. It should work also when browsing with the keyboard (as it does for the paramers in the composite model panel).
  * Ranges are not supported. Neither are convenient operations like "Freeze/Thaw all, Reset, etc.", which are provided in Iris 2.1 but have no requirements in the BRD.
  * Units are set to Angstrom vs erg/s/cm2/Angstrom.
  * It would be nice to have a more flexible and good-looking notification area, as a widget to embed in each frame or as a single notification area on the desktop where components can enqueue infos, warnings, and errors.

There is a number of other small changes made to either improve testing (I finally found out how to discover desktop buttons, i.e. labels with UISpec4J! They are considered TextBoxes!), clean up various oddities or remove unused classes, or simply the result of refactoring, including class name changes and so on.


samp-factory/src/test/java/cfa/vo/sherpa/models/ModelFactoryTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertEquals((Double) (-10.0), par.getMin());
+        assertEquals((Double) 10.0, par.getMax());
+//        assertEquals("", par.getUnits());
+//        assertEquals("INDEF", par.getLink());
+    }
+
+    @Test(expected=NoSuchElementException.class)
+    public void testParameterException() throws Exception {
+        ModelImpl m = factory.getModel("powerlaw1d", "p1d");
+        m.getParameter("foo");
```
jbudynk: Should be `powerlaw` instead of `powerlaw1d`. The test never gets to this line because the line above it can't find the model and errors with a NoSuchElementException.

samp-factory/src/main/java/cfa/vo/sherpa/models/ModelFactory.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    private void build(URL modelsFile) {
+        try {
+            DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+            DocumentBuilder builder = factory.newDocumentBuilder();
+            Document doc = builder.parse(modelsFile.openStream());
+            Element root = doc.getDocumentElement();
+            NodeList models = root.getElementsByTagName("model");
+            for (int i=0; i<models.getLength(); i++) {
```
eholum: Any reason to put the try catch inside the for-loop? What happens if there is only one invalid model in the xml file?

samp-factory/src/main/java/cfa/vo/sherpa/SherpaClient.java
```diff
@@ -24,8 +28,9 @@
 
 public class SherpaClient {
 
+    private ModelFactory modelFactory = new ModelFactory();
     private SampService sampService;
-    private Map<String, AbstractModel> modelMap = new HashMap<>();
+    private Map<String, ModelImpl> modelMap = new HashMap<>();
```
eholum: Should this depend on the implementation or the interface?

samp-factory/src/main/java/cfa/vo/sherpa/models/ModelImpl.java
```diff
@@ -0,0 +1,120 @@
...
+
+    private Map<String, Parameter> parMap = new HashMap();
+
+    private String name;
+
+    private String description;
+
+    private String id;
+
+    public ModelImpl(String name, String id) {
```
eholum: Maybe make these constructors protected, as it looks like they're only built through the factory class.

eholum:
Looks good to me! Just some minor comments above - otherwise I think it's good to merge.

samp-factory/src/main/java/cfa/vo/sherpa/models/ModelFactory.java
olaurino: If there are issues in the xml I want to know and fix the file. This is a static file distributed with Iris and rarely updated. Not something we get from an external service (although I always wandered whether it should be sherpa-samp to serve it, but that's a different discussion).

samp-factory/src/main/java/cfa/vo/sherpa/SherpaClient.java
olaurino: I left this for some final refactoring. The reason I am using the implementation is that there is a method in it that follows the javabean accessor convention even though it's not a real accessor, and it is not even included in the interface. I'll probably end up changing the name of the method and pull it up in the interface.

olaurino:
I pushed to this branch rather than opening a new PR. If any of you @eholum and @jbudynk feel I should rather open a new PR that's fine with me, that was the original plan.

In any case you can start reviewing the work for chandracxc/iris-dev#16
I still need to include custom models in the picture, but PRs in java tend to explode easily, so if you had time to review the work so far that would be good.

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingMainView.java
```diff
@@ -0,0 +1,526 @@
...
+import javax.swing.*;
+import javax.swing.tree.*;
+import java.awt.GridLayout;
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Logger;
+
+public class FittingMainView extends javax.swing.JInternalFrame implements SedListener {
```
eholum: It might be worth adding a layout manager (GridBagLayout or something) in the early stages. To support keeping things in order on resizes, etc.

iris-common/src/main/java/cfa/vo/iris/fitting/FitConfigurationBean.java
```diff
@@ -0,0 +1,327 @@
...
+
+        setrStat(results.getRstat());
+        setnFev(results.getNfev());
+        setqVal(results.getQval());
+        setNumPoints(results.getNumpoints());
+        setStatVal(results.getStatval());
+        setDof(results.getDof().intValue());
+    }
+
+    public SherpaFitConfiguration make(ExtSed sed) throws SedException, UnitsException {
```
eholum: Would probably be worth having a constructor that just takes double[] of spectral, flux, and flux error values directly. Fit configurations don't have to be tied to ExtSeds

iris-common/src/main/java/cfa/vo/iris/fitting/FitConfigurationBean.java
```diff
@@ -0,0 +1,327 @@
...
+    {
+        propertyChangeSupport.removePropertyChangeListener( listener );
+    }
+
+    private void removeUserModel(Model m) {
+        boolean retVal = false;
+        for (UserModel um : new ArrayList<>(userModelList)) {
+            String mId = getId(m);
+            String umId = getId(um);
+            if (mId.equals(umId)) {
```
eholum: Use StringUtils for string operations - it's good practice and null safe.

iris-common/src/main/java/cfa/vo/iris/fitting/FitConfigurationBean.java
```diff
@@ -0,0 +1,327 @@
...
+        return stripId(m.getName());
+    }
+
+    private String stripId(String name) {
+        return name.split("\\.")[1];
+    }
+
+    private void addToExpression(String id) {
+        String expression = model.getName();
+        if (expression == null || expression.isEmpty()) {
```
eholum: Ditto, StringUtils.isEmpty or isBlank

iris-common/src/main/java/cfa/vo/iris/fitting/custom/DefaultCustomModel.java
```diff
@@ -0,0 +1,349 @@
...
+        for (int i=0; i<parNames.length; i++) {
+            try {
+                Parameter p = SAMPFactory.get(Parameter.class);
+                p.setName(parNames[i]);
+                p.setVal(parVals[i]);
+                p.setMin(parMins[i]);
+                p.setMax(parMaxs[i]);
+                p.setFrozen(parFrozens[i] ? 1 : 0);
+                m.addPar(p);
+            } catch(IndexOutOfBoundsException ex) {
```
eholum: Check this condition before starting the loop. ifs are better than try/catch.

olaurino: In general I would agree, but the fact that each array might, in principle, have a different length, makes the `if` particularly ugly. On the other hand, the custom model manager does not currently provide any feedback to the user regarding their input. This is a known issue, and it should be properly addressed before the custom models are instantiated. This class and the whole custom model manager requires some serious reworking if time allows [btw I found the email thread where it shows that support for custom models was coded in 3 days].

olaurino:
Force-pushing a number of commits to trigger the travis builds. I'll keep doing some clean up and address some issues. I'll send a notification when this is ready to be reviewed. Sorry I didn't manage to keep the changes more atomic!

Original tip was at 51ac268.

iris-common/src/main/java/cfa/vo/iris/fitting/FitConfigurationBean.java
olaurino: Agreed.

olaurino: Hold on, I am not sure I agree. I think I would prefer to keep the Sed abstraction, but we can discuss about this.

olaurino: Let me clarify: currently our fits are rather simple: just pass x, y, statErr. However, one may add support for systematic errors, asymmetric error bars, upper/lower limits and so on. We have these in the original list of requirements. Moreover, an aggregated SED is a complex object that may contain different kinds of segments with different kinds of treatments when they are included in a fit, e.g. one may decide to resample HD spectra or to include priors and what not.

So, both in principle and in practice, it is an SED as a complex object that needs to be fitted. The fact that currently we only flatten the SED and take x, y, and err, is a limitation, and does not capture the breadth of the original requirements.

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingMainView.java
olaurino: I experimented a little bit with layouts but it was distracting, and I managed to get good results with free design (the only issue when running the GUI on Mac is that the size of the starting window is a little bit smaller than it should, but I think I know how to fix that.

I would like to add layers to make the GUI look more stable, nevertheless.

jbudynk:
First off, the new model chooser looks great! The model descriptions shown at the bottom of the panel is a great addition. I also like that you include "1.6" as the default sigma option, with the % next to it.

some GUI issues:

 - the "Fit Succeeded" box is cut-off by the fit results panel.
 - when first opening the Fitting Tool, the bottom part of the frame is cut-off.
 - after a fit, if any model was expanded to view the model parameters (in the Model Components tree), those tabs are closed if opened prior.

bugs:
 - I can't compute the confidence intervals more than once (though I'll need to test that some more...)
 - the model search bar doesn't work for custom models (but that's understandable considering the custom model component needs some work)

This is a huge improvement over the old fitting tool - nice work!

iris-common/src/main/java/cfa/vo/iris/sed/ExtSed.java
```diff
@@ -59,6 +61,18 @@
     private Map<String, Object> attachments = new TreeMap();
     private String id;
     private boolean managed = true;
+
+    @Nonnull
+    public FitConfiguration getFit() {
+        return fit;
+    }
+
+    public void setFit(FitConfiguration fit) {
```
eholum: This needs a null check on it for that annotation.

iris-common/src/main/java/cfa/vo/iris/gui/widgets/ModelExpressionVerifier.java
```diff
@@ -0,0 +1,54 @@
...
+        if (models != null) {
+            for (Model m : models) {
+                b.put(DefaultModel.findId(m), Double.NaN);
+            }
+        }
+
+        try {
+            scriptEngine.setBindings(b, ScriptContext.ENGINE_SCOPE);
+            Object res = scriptEngine.eval(expression);
+            if (res != null) {
```
eholum: nbd, but you can remove the null check by just checking against the known static constant.

Double.NaN.equals(res)

iris-common/src/main/java/cfa/vo/sherpa/SherpaClient.java
```diff
@@ -0,0 +1,166 @@
...
+import cfa.vo.interop.*;
+import cfa.vo.iris.fitting.FitConfiguration;
+import cfa.vo.iris.sed.ExtSed;
+import cfa.vo.iris.utils.UTYPE;
+import cfa.vo.sherpa.models.*;
+import org.astrogrid.samp.Response;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Logger;
+
+public class SherpaClient {
```
eholum: So much better to have :)

iris-common/src/test/java/cfa/vo/sherpa/SherpaClientTest.java
```diff
@@ -0,0 +1,82 @@
...
+    SherpaClient client;
+    private CompositeModel cm;
+    double[] x = {1.0, 2.0, 3.0};
+    double[] y = {1.0, 4.0, 9.0};
+
+    @Before
+    public void setUp() throws Exception {
+        client = new SherpaClient(Mockito.mock(SampService.class));
+        sed = new ExtSed("testsed", false);
+        Segment segment = new Segment();
```
eholum: There's a common method in testutils that does exactly this. TestsUtils.createSampleSegment or something...

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingMainView.java
```diff
@@ -0,0 +1,489 @@
...
+        }
+    }
+
+    private class AvailableModelsMouseAdapter extends MouseAdapter {
+        @Override
+        public void mousePressed(MouseEvent e) {
+            TreePath selPath = availableTree.getPathForLocation(e.getX(), e.getY());
+            if (selPath != null) {
+                DefaultMutableTreeNode node = (DefaultMutableTreeNode) selPath.getLastPathComponent();
+                if (node.isLeaf()) {
```
eholum: The 'tables' and 'functions' options under user models aren't really leafs, so this is throwing an exception. But those are eventually going to be folders, right?

eholum:
Nice, ditto to what Jamie said. Huge improvement over the old version. Same notes on the GUI for the fitting tool - I think the default window size just needs to be a little bigger.. 

Only other issue is that every now and then running the fitting tool on my mac causes a sherpa exception ... Says it can't connect to sherpa despite the icon saying it's connected.

 Other than than just minor comments, I'm a fan of the model/controller view setup going on

I think we should merge this (to close it out), knowing that there are still a bunch of known issues. They will continue to be fixed as the development for the fitting tool continues.

olaurino:
7bf07f4 should address the window size issue on OSX reported in the PR description as well as in https://github.com/ChandraCXC/iris/pull/267#issuecomment-211602004. @jbudynk are all the issues in this comment still relevant?

iris-common/src/main/java/cfa/vo/iris/sed/ExtSed.java
olaurino: It was guaranteed to be non null at some point before the refactoring. Commit 9de3708 should fix it. I was hoping to get more IDE support with these annotations, but the fact that this issue fell through the cracks indicates such support is limited (at least in IDEA). If we start using them we should indeed pay attention that they are correctly enforced.

iris-common/src/main/java/cfa/vo/iris/gui/widgets/ModelExpressionVerifier.java
olaurino: There is no Double.NaN.equals method. I could use Double.isNaN(obj) but that requires obj to be a Double itself, which might be not be applicable.

iris-common/src/test/java/cfa/vo/sherpa/SherpaClientTest.java
olaurino: Let's make it a habit to document this kind of stuff in the UDF. There is a section for this. Not sure I kept it up to date either, though :)

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingMainView.java
olaurino: Good catch. Not sure whether it's worth fixing, though... it does not impact functionality, unless we decide we want to improve error handling and always show users an error message when the casting fails. If that's the case, I would create a new task for it and merge this PR without this change. Or am I missing something?

olaurino:
Travis is green. Can I merge this?

eholum:
Merge it, though are you going to merge into sprint-9 or into master? My preference would be the former.

olaurino:
Why sprint-9 rather than master?

olaurino:
Note that this cannot be merged in sprint-9 as it was issued against master (maybe it should have been sprint-8, but that would have had its drawbacks too). So I would need to issue a new PR based on the same branch, in order not to lose the GitHub tracking.

[BTW, I started to like that we merged against sprint PRs, but as time goes by I think I changed my mind... that's irrelevant in this discussion, though.]

eholum:
We've just been re-opening PRs against the current sprint before merging. Merging into a sprint branch has been in our working agreements (with positive feedback about the benefits it gives us at our retrospectives) since sprint-6.

olaurino:
Sounds good.

PR #269, opened 2016-03-21
iris-dev#31 Filtering Points in the Metadata Browser

eholum:
This PR relies heavily on https://github.com/ChandraCXC/iris/pull/266, but I would rather split them up. I would ask that https://github.com/ChandraCXC/iris/pull/266 to be reviewed and merged first before digging into this one.

Closes

https://github.com/ChandraCXC/iris-dev/issues/31
https://github.com/ChandraCXC/iris-dev/issues/32

Additional minor changes were required in some of the ColumnInfoMatching logic in order to allow custom settings of default values for non-overlapping columns. And SegmentStarTables column logic has been improved and organized to support the new 'Filter Column'.


eholum:
Closing this PR. Will re-open one against the sprint 8 branch.

PR #270, opened 2016-03-25
iris-dev-13: implement plot preferences

jbudynk:
This PR was originally ChandraCXC/iris#262, but is now being compared to sprint-7 instead of sprint-6.

### Changes:

- migrated the PlotterView to Netbeans
- made PlotPreferences a SED-specific object, meaning each SED has their own set of preferences.
- added a fixed-view capability, which when checked (menu: View --> Fixed) prevents the plot ranges from updating when SED/Segment changes occur
- plot types (linear, log, xlog, and ylog) are implemented. The plot type is retained when user switches between SEDs.
- the grid lines can be turned on and off.
- functional tests for these preferences have been added to VisualizerComponentTest.
- StilPlotter is a bean now. This was done for the PlotterView-to-Netbeans migration.
- The VisualizerComponentPreferences property in StilPlotter is no longer final.

I tried adding zoom in/out functionality to the PlotterView "In" and "Out" buttons, but I couldn't come up with a working zoom algorithm, so these buttons are now disabled.

### TODOs:

The TODOs are beyond Tier 1 tasks.

- implement show/hide error bars functionality
- come up with a preferences GUI widget where the user can set truly global plot preferences that apply to all the SEDs, or to the selected SEDs.


iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/OtherPlotPreferences.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import uk.ac.starlink.ttools.plot2.geom.PlaneAspect;
+import uk.ac.starlink.ttools.plot2.task.PlanePlot2Task;
+
+/*
+ * These are plot preferences outside the STILTS regime.
+ */
+public class OtherPlotPreferences {
```
eholum: I'm not sure that these merit their own class - just leave them in PlotPreferences and add comments for things that are stil related and not, or if things aren't stil related make them class variables rather than putting them in the map. That would probably be clear enough, and It keeps everything related to user settings in a single place.

If we feel like the plot preferences class is too complicated, then we should put specific preferences inside smaller beans. Inside the PlotPreferences class.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
```diff
@@ -129,12 +82,17 @@ public PlotterView(String title,
...
         
-        initializeComponents();
-        initializeMenuItems();
+        initComponents();
+        
+        // initializing the stil plotter
+        stilPlotter1.setWorkSpace(ws);
+        stilPlotter1.setSedManager((SedlibSedManager) ws.getSedManager());
+        stilPlotter1.setVisualizerPreferences(preferences);
+        this.plotter = stilPlotter1;
```
eholum: Just reference stilPlotter1 rather than plotter. O/w you have two class variable pointing to the same object.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
```diff
@@ -193,201 +198,452 @@ private void redrawPlot() {
...
+    private javax.swing.JToggleButton tglbtnShowHideResiduals;
+    private javax.swing.JTextField txtXposition;
+    private javax.swing.JTextField txtYposition;
+    private cfa.vo.iris.gui.JButtonArrow up;
+    private javax.swing.JButton zoomIn;
+    private javax.swing.JButton zoomOut;
+    private org.jdesktop.beansbinding.BindingGroup bindingGroup;
+    // End of variables declaration//GEN-END:variables
+
+    private void makeLinear() {
```
eholum: These 4 methods could be one method that takes a PlotType as a param.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
```diff
@@ -193,201 +198,452 @@ private void redrawPlot() {
...
+     * @param fixed set to "true" to fix the viewport when SED changes occur, 
+     * and set to "false" to let the viewport resize automatically with changes.
+     */
+    private void setFixedViewPort(boolean fixed) {
+        try {
+            this.preferences.getSelectedSedPreferences().getPlotPreferences().setFixed(fixed);
+            
+            // needs to be set whenever viewport changes
+            this.preferences.getSelectedSedPreferences().getOtherPlotPreferences().setAspect(this.plotter.getPlotDisplay().getAspect());
+        } catch (UnsupportedOperationException ex) {
```
eholum: What's throwing this exception?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java
```diff
@@ -82,7 +98,16 @@ public StilPlotter(IWorkspace ws,
      *            current display is active.
      */
     public void reset(ExtSed sed, boolean dataMayChange) {
-        resetPlot(sed, false, dataMayChange);
+        boolean fixed;
+        try {
+            fixed = this.getVisualizerPreferences()
+                    .getSelectedSedPreferences().getPlotPreferences().getFixed();
+        } catch (NullPointerException ex) {
```
eholum: It's generally better practice to just add a null check where appropriate than rely on an exception being thrown. If the user opens the plotter and no SED is selected, then shouldn't sed be null?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java
```diff
@@ -82,7 +98,16 @@ public StilPlotter(IWorkspace ws,
      *            current display is active.
      */
     public void reset(ExtSed sed, boolean dataMayChange) {
-        resetPlot(sed, false, dataMayChange);
+        boolean fixed;
+        try {
+            fixed = this.getVisualizerPreferences()
```
eholum: Just thinking more about this, we should probably be relying on the specified sed in each one of these different components, rather than calling getSelectedSedPreferences - which is inherently not thread safe given that it's updated by the swing EDT.

Where possible call getSetPreferences(selectedSed) rather than getSelectedSedPreferences

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java
```diff
@@ -143,7 +177,115 @@ private void resetPlot(ExtSed sed,
...
+    /**
+     * Zoom in or out of plot by a given scale factor.
+     * @param zoomFactor 
+     */
+    public void zoom(double zoomFactor) {
+        
+        throw new UnsupportedOperationException("The zoom functionality is"
+                + "currently unsupported.");
 
+        // TODO: this algorithm is BAD! Need to implement a better one.
```
eholum: I looked more into Stils stuff, what we need should be somewhere in https://github.com/Starlink/starjava/tree/83a1eef2bb85ff87772b688803cec44f98745665/ttools/src/main/uk/ac/starlink/ttools/plot2/geom

Specifically in the PlaneNavigator, but haven't dug too deeply yet.

Something like:

PlaneSurface s = (PlaneSurface) this.getPlotDisplay().getSurface();
s.getAxes()[0].dataZoom(arg0, arg1)

should work, but i don't know about the zoom factors. It would also allow us to easily fix x and y axis on the zoom buttons.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/OtherPlotPreferences.java
jbudynk: Ah, I should've documented this.   I agree that it's weird to have another class just with the Aspect defined, but there's a reason. I'm sure there's a better solution than the hack one I came up with :)

Say a user opens a SED, zooms in, and selects "Fixed" from the menu. Looking at `setFixedViewPort()` ([PlotterView line 586](https://github.com/ChandraCXC/iris/pull/270/files#diff-d16d4d6a7d7ec5844bfa47dedeac896eR586)), any time the "Fixed Plot" menu item is checked, the Aspect is saved. In `setupMapEnvironment()` in [StilPlotter at line 365](https://github.com/ChandraCXC/iris/pull/270/files#diff-62c68282ac1c34f74a07fa31fb632756R365), everything in the PlotPreferences object is copied into the MapEnvironment used to create the PlotDisplay. If the user clicks the "Reset" button, the SED is temporarily unfixed (see [`btnResetActionPerformed`](https://github.com/ChandraCXC/iris/pull/270/files#diff-d16d4d6a7d7ec5844bfa47dedeac896eR501)  in PlotterView), so that the plotter will be reset to the full data range. `btnResetActionPerformed` eventually calls `createPlotComponent()`, which calls [`setupMapEnvironment()`](https://github.com/ChandraCXC/iris/pull/270/files#diff-62c68282ac1c34f74a07fa31fb632756R354), which then grabs the PlotPreferences, and sets the PlotDisplay environment to whatever the plot preferences are. The Aspect was set in `setFixedViewPort()`, and so clicking the "Reset" button won't reset to the full plot range.

But if I don't store the Aspect at all, then keeping the latest fixed viewport for other SED changes (adding/removing segments, switching SEDs) is impossible. Therefore, I store the Aspect in another class so the Reset button works properly.

This is bad code, and I want to refactor it to make it better. But this works for now. I'll definitely take a look back at this later, but for now, I don't have time. If you have any suggestions, let me know - I'm all ears!

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
jbudynk: if the user opens the Plotter before a SED is loaded in, the selectedSed is undefined. The VisualizerComponentPreferences object in PlotterView has a default PlotPreferences object that is global, so this is to catch that in case the user clicks "Fixed" from the menu.

... and now I see that's an UnsupportedOperationException, not a NullPointerException... The Unsupported exception is old. At one point, fixed-view port was not implemented, so I added this try/catch block.

I'll replace this with a NullPointerException:

```
} catch (NullPointerException ex) {
    this.preferences.getPlotPreferences().setFixed(fixed);
    this.preferences..getOtherPlotPreferences().setAspect(this.plotter.getPlotDisplay().getAspect());
}

File: iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java (hunk 5 above)
Comment (jbudynk) [inline iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java]: I see. How about

```
if (sed == null) {
    // statement under the 'catch' block
} else {
    // statement under the 'try' block
}

eholum: Yep.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/OtherPlotPreferences.java
eholum: Ah, so the aspect needs to be attached directly to the SED rather than in the preferences object. So what about just putting it as a class variable in the sedpreferences object? Or put it in plot preferences as a class variable, and just don't put it in the relevant map?

jbudynk:
I've addressed @eholum 's comments (besides the ones remaining.)

- I replaced all `getSelectedSedPreferences()` with `getSedPreferences(sed)` so that we're looking at the same SED in all objects that look for the selected SED.
- I've added a legend that isn't cut-off. Right now, it just has default options that can't be accessed by the user.
- I switched a lot of `try/catch` blocks to `if/else` statements so that it's more obvious what the conditions are, and so that we're "catching" the right conditions, and not covering over an independent exception.

### TODOs:
- refactor the OtherPlotPreferences class back into PlotPreferences
- make it so that when one switches grid on/off, and the plot is fixed, that the plot doesn't reset.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
```diff
@@ -147,11 +104,58 @@ public void actionPerformed(ActionEvent e) {
...
+                mntmYlog.setSelected(true);
+                changePlotType(PlotPreferences.PlotType.Y_LOG);
+            }
+        });
+        
+        // Action to toggle grid on/off
+        mntmGridOnOff.addActionListener(new ActionListener() {
+            @Override
+            public void actionPerformed(ActionEvent e) {
+                boolean on = !PlotterView.this.plotter.getVisualizerPreferences().getSedPreferences(plotter.getSed()).getPlotPreferences().getShowGrid();
```
eholum: Could this just be mntmGridOnOff.isSelected() or whatever? Would that shorten it?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
```diff
@@ -147,11 +104,58 @@ public void actionPerformed(ActionEvent e) {
...
+                mntmGridOnOff.setSelected(on);
+                setGridOn(on);
+            }
+        });
+        
+        // Action to fix or unfix the plot viewport
+        mntmAutoFixed.addActionListener(new ActionListener() {
+            @Override
+            public void actionPerformed(ActionEvent e) {
+                boolean fixed = !PlotterView.this.plotter.getVisualizerPreferences().getSedPreferences(plotter.getSed()).getPlotPreferences().getFixed();
```
eholum: Ditto this

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java
```diff
@@ -120,9 +151,19 @@ private void resetPlot(ExtSed sed,
         if (display != null) {
             display.removeAll();
             remove(display);
-            existingAspect = display.getAspect();
+            
+            if (this.preferences.getSedPreferences(currentSed) != null) {
```
eholum: I don't think you need this anymore, calling getPlotPreferences() will ensure that the current plotpreferences is not null, then can't you just call

getPlotPreferences().setAspect(display.getAspect());

Whenever the display is not null? That would save the existing aspect in the current SED's plot preferences every time reset plot is called.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java
```diff
@@ -133,9 +174,16 @@ private void resetPlot(ExtSed sed,
...
         display = createPlotComponent(currentSed, cached);
         
-        // Set the bounds using the aspect if provided one and if the plot is fixed
-        if (existingAspect != null && fixed) {
-            display.setAspect(existingAspect);
+        // Set the bounds using the aspect if the plot is fixed
+        if (fixed) {
+            // By now, the current SED has switched from the previous one to 
+            // the current one. Update the aspect to the currently-selected SED
+            if (this.preferences.getSedPreferences(currentSed) != null) {
```
eholum: Same deal, using the existing getPlotPreferences() means you would never have to check here.

Though it would fix the 'empty' sed plot. But I don't think that's a problem?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java
```diff
@@ -143,14 +191,116 @@ private void resetPlot(ExtSed sed,
...
+    /**
+     * Change the plotting space between logarithmic and linear. One of the axes
+     * can be logarithmic, while the other is linear.
+     * @param plotType the plot type to use. Must be one of the 
+     * PlotPreferences.PlotType enums. Choices are LOG, LINEAR, XLOG, and YLOG.
+     */
+    public void changePlotType(PlotPreferences.PlotType plotType) {
+        
+        try {
+            preferences.getSedPreferences(currentSed).getPlotPreferences().setPlotType(plotType);
```
eholum: Use built in method

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java
```diff
@@ -143,14 +191,116 @@ private void resetPlot(ExtSed sed,
...
+        
+        updatePlotDisplay();
+        
+    }
+    
+    public void setGridOn(boolean on) {
+        setupForPlotDisplayChange();
+        
+        try {
+            preferences.getSedPreferences(currentSed).getPlotPreferences().setShowGrid(on);
```
eholum: Ditto

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPreferences.java
```diff
@@ -30,16 +31,64 @@
...
+    public static final String PLOT_TYPE = "plot_type"; // not STILTS
+    //public static final String SHOW_ERRORS = "show_errors"; // not STILTS
+    
+    // for the plot legend
+    public static final String SHOW_LEGEND = "legend";
+    public static final String LEGEND_BORDER = "legborder";
+    public static final String LEGEND_OPAQUE = "legopaque";
+    public static final String LEGEND_POSITION = "legpos";
+    
+    private PlaneAspect aspect; // not STILTS. Do not add to the prefs map!
```
eholum: Nice! I think this is a lot cleaner :)

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
jbudynk: Done.

I still have an issue where the first time you open the Plotter -- whether a SED is loaded or not -- the 'Grid on/off' checkbox is always unchecked. I have the checkbox bound to `VisualizerComponentPreferences.plotPreferences.showGrid` which is `true` by default, yet that check box is never checked the first time the Plotter is opened. Any idea why?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java
jbudynk: > Whenever the display is not null? That would save the existing aspect in the current SED's plot preferences every time reset plot is called.

This is true, but the plotter will never set the display to the stored aspect, unless the plot is fixed. I see how this could cause problems down the road if we change implementation enough. I'll leave a comment for now, but if I change the conditional to `if (fixed)`, I won't be able to reset the plot to the full data range.

... I'll see if I can clean this function up some more...

jbudynk:
Thanks for the latest batch of feedback! I've added in the recommended changes:

- using StilPlotter's getPlotPreferences() built in method for getting the correct plot preferences out.
   - I also made another getPlotPreferences() that accepts an ExtSed. This was for getting the `fixed` property for the correct SED (see `reset()` at line 97 in StilPlotter).
- `StilPlotter::resetPlot()` has been greatly refactored and simplified.
- the grid and fixed preferences actionPerformed() methods have been updated as suggested.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/StilPlotter.java
```diff
@@ -252,4 +347,42 @@ private void addSegmentLayers(ExtSed sed, MapEnvironment env)
...
+    
+    public PlotPreferences getPlotPreferences() {
+        if (this.preferences.getSedPreferences(currentSed) != null) {
+            return preferences.getSedPreferences(currentSed).getPlotPreferences();
+        } else {
+            return preferences.getPlotPreferences();
+        }
+    }
+    
+    public PlotPreferences getPlotPreferences(ExtSed sed) {
```
eholum: Nice, yeah I like this, and ultimately down the road when we support co-plotting I think it'll be super nice to have.

eholum:
Looks good to me!

Re: the grid issue when there's no SED in the plotter - I'm not sure where exactly this is happening, but an interim easy fix is just to set it manually to true in the constructor. Not sure why that binding isn't working but it'll probably be changed if/when we cleanup the preferences infrastructure again.

For now I think this looks good. 

The travis builds failed on the sherpa exception, but I just ran them twice on my machine and they were fine.

Can you rebase and merge it into sprint-7?

jbudynk:
Rebased on latest sprint-7 updates. Was based on 0cee6a4 with the tip 2d43d5325. 

This affects branch 34-mouse-listeners (ChandraCXC/iris-dev#34), as I see some of it was based on the work in this PR.

I also added a very hackish fix for the grid and fixed-plot preferences when opening the first SED. I made note that the plot preferences should fire property changes to the the PlotterView and StilPlotter so that we can use beans bindings properly (but I just don't have the time to make the change at the moment).

eholum:
I think the bindings will be easier to manage once the model is a little cleaner between the visualizer components. For now let's merge it!

jbudynk:
rebased again on latest sprint-7 changes. Original tip was ed209ac.

PR #272, opened 2016-03-30
iris-dev#31 Applying Masks to StarTables

eholum:
Closes

ChandraCXC/iris-dev#31
ChandraCXC/iris-dev#32

Re-opening against sprint-8 branch. Replaces: https://github.com/ChandraCXC/iris/pull/269

Additional minor changes were required in some of the ColumnInfoMatching logic in order to allow custom settings of default values for non-overlapping columns. And SegmentStarTables column logic has been improved and organized to support the new 'Filter Column' - which is displayed iff there is at least one filtered row in the selected tables.


eholum:
As per what I learned yesterday in our planning meeting, I changed the names of the filter classes to 'mask' to make it more clear that they are actually 'MASKING' points, not just a filter for selection.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/filters/NullMask.java
```diff
@@ -19,7 +19,7 @@
 
 import cfa.vo.iris.visualizer.stil.tables.IrisStarTable;
 
-public class NullFilter implements Filter {
+public class NullMask implements Mask {
     
     @Override
     public BitSet getFilteredRows(IrisStarTable table) {
```
jbudynk: What's the point of this class? all 'getFilteredRows()' does is return an empty BitSet. Is that intended?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/filters/Mask.java
```diff
@@ -18,7 +18,7 @@
 import java.util.BitSet;
 import cfa.vo.iris.visualizer.stil.tables.IrisStarTable;
 
-public interface Filter {
+public interface Mask {
     
     public abstract BitSet getFilteredRows(IrisStarTable table);
```
jbudynk: If this method is meant to return the *masked* points, I'd call it "getMaskedRows" instead of "getFilteredRows."

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/filters/NullMask.java
eholum: Yeah it's intended, at some point I was thinking that a mask could act like a filter - and we may have different kinds of masks that act on different points. This was a legacy of the simplest kind of mask (one that does nothing). The other mask implementation that's currently in here is based on rows of a star table. Would we ever want one that is based on star table values (like mask all points with characterization x)?

That being said, this was put in to differentiate between an actual empty mask vs a row subset mask with no values, but is not used in live code anymore, it's rather a marker that the mask type is extensible. If we don't want it to be extensible then I'd rather take the abstract class out and just leave masks as a pojo.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/filters/Mask.java
eholum: Excellent point, I'll change those and add it to the last commit.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/filters/NullMask.java
jbudynk: Ah, ok. Leave it in then.

> "Would we ever want one that is based on star table values (like mask all
points with characterization x)?"

I'd say yes. That's what I imagine a "don't show points without errors"
filter to be like (and I say "filter" because we used a filter to select
the points to mask out in Iris 2.1).  I remember having a conversation to
include a kind of "pre-built" filter that masks points without errors. I
think this falls under Plotter task 3.1.a in the BRD.

On Wed, Mar 30, 2016 at 4:10 PM, Erik Holum <notifications@github.com>
wrote:

> In
> iris-visualizer/src/main/java/cfa/vo/iris/visualizer/filters/NullMask.java
> <https://github.com/ChandraCXC/iris/pull/272#discussion_r57954998>:
>
> > @@ -19,7 +19,7 @@
> >
> >  import cfa.vo.iris.visualizer.stil.tables.IrisStarTable;
> >
> > -public class NullFilter implements Filter {
> > +public class NullMask implements Mask {
> >
> >      @Override
> >      public BitSet getFilteredRows(IrisStarTable table) {
>
> Yeah it's intended, at some point I was thinking that a mask could act
> like a filter - and we may have different kinds of masks that act on
> different points. This was a legacy of the simplest kind of mask (one that
> does nothing). The other mask implementation that's currently in here is
> based on rows of a star table. Would we ever want one that is based on star
> table values (like mask all points with characterization x)?
>
> That being said, this was put in to differentiate between an actual empty
> mask vs a row subset mask with no values, but is not used in live code
> anymore, it's rather a marker that the mask type is extensible. If we don't
> want it to be extensible then I'd rather take the abstract class out and
> just leave masks as a pojo.
>
> —
> You are receiving this because you commented.
> <https://github.com/ChandraCXC/iris/pull/272/files/9235c774d437d4dfda6062c44bf57d79e5bf983e..434887e6b251bd33aeab4ff4a1fde0c165038b63#r57954998>
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentColumn.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        @Override
+        public Object readValue(long arg0) throws IOException {
+            return name;
+        }
+    }
+    
+    /**
+     * Similar to a ConstantColumn, but with an adjustable name value.
```
eholum: Whoops, fixed this comment

eholum:
Added additional information on comments, etc...

PR #273, opened 2016-03-30
iris-dev#15 Several Mouse Function for the Plotter

eholum:
Small PR which will setup later tasks for other mouse listeners.

For now, this closes: https://github.com/ChandraCXC/iris-dev/issues/15, and will display the mouse's coordinates in the stil plotter in the two places we reserved for them in the plot view.

I figure for now we can put mouse listener classes into the MouseListenerManager class, which the StilPlotter can maintain a pointer to to make sure each new mouse listener is added to the PlotDisplay whenever one is created. I left it in the preferences class so that other callers have access if necessary.

There's also a simple, silly Point Selection Listener in place just to show how it works. Whomever picks up the point selection event tasks can use it as a template for what point selection events look like.

[EDIT]

This now also closes 
https://github.com/ChandraCXC/iris-dev/issues/28
https://github.com/ChandraCXC/iris-dev/issues/29

As mentioned in TODOS/comments, there is going to be more work when we implement sorting and after we merge the masking PR (https://github.com/ChandraCXC/iris/pull/272). I think the separation is pretty clear in the code, and that this won't have to change when those are implemented.

Also note that point selection in the PlotDisplay VIA hovering is pretty limited, it would be nice if the PlotDisplay class supported calling https://github.com/Starlink/starjava/blob/master/ttools/src/main/uk/ac/starlink/ttools/plot2/task/PlotDisplay.java#L405 or had some other way get get at nearest points other than the single interface. I added a question for them in the UDF. For now the selection/metadata listeners preserve existing Iris2.1 behaviour.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2016 Chandra X-Ray Observatory.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package cfa.vo.iris.visualizer.plotter;
+
+import java.awt.Point;
+import java.awt.event.MouseEvent;
+import java.awt.event.MouseMotionListener;
+import java.text.DecimalFormat;
+
+import cfa.vo.iris.visualizer.stil.StilPlotter;
+import uk.ac.starlink.ttools.plot2.Surface;
+import uk.ac.starlink.ttools.plot2.task.PlotDisplay;
+
+public class MouseCoordinateMotionListener extends StilPlotterMouseListener 
+    implements MouseMotionListener {
+    
+    private PlotDisplay display;
+    
+    private final static DecimalFormat df = new DecimalFormat("0.######E0");
+    static {
+        df.setMaximumFractionDigits(6);
+    }
+
+    @Override
+    public void setPlotterView(PlotterView plotterView) {
+        super.setPlotterView(plotterView);
+    }
+
+    @Override
+    public void activate(PlotDisplay display) {
+        display.addMouseMotionListener(this);
+    }
+
+    @Override
+    public void mouseDragged(MouseEvent evt) {
+    }
+
+    @Override
+    public void mouseMoved(MouseEvent evt) {
+        if (display == null || display.getSurface() == null) {
+            return;
+        }
+
+        Surface surface = display.getSurface();
+
+        Point p = evt.getPoint();
+        if (!surface.getPlotBounds().contains(p)) {
+            return;
+        }
+
+        double[] loc = surface.graphicsToData(p, null);
+        plotterView.setXcoord(formatNumber(loc[0]));
+        plotterView.setYcoord(formatNumber(loc[1]));
+    }
+
+    private static String formatNumber(double d) {
+        return df.format(d);
+    }
+}
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
@@ -41,6 +41,7 @@ public class MouseCoordinateMotionListener extends StilPlotterMouseListener
 
     @Override
     public void activate(PlotDisplay display) {
+        this.display = display;
         display.addMouseMotionListener(this);
     }
 
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
@@ -22,6 +22,11 @@ import java.awt.event.MouseMotionListener;
 import uk.ac.starlink.ttools.plot2.Surface;
 import uk.ac.starlink.ttools.plot2.task.PlotDisplay;
 
+/**
+ * MouseListener that displays the mouse's current coordinates (in the PlotDisplay's
+ * coordinate system) in the relevant PlotterView text objects.
+ *
+ */
 public class MouseCoordinateMotionListener extends StilPlotterMouseListener 
     implements MouseMotionListener {
     
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
@@ -18,21 +18,14 @@ package cfa.vo.iris.visualizer.plotter;
 import java.awt.Point;
 import java.awt.event.MouseEvent;
 import java.awt.event.MouseMotionListener;
-import java.text.DecimalFormat;
 
-import cfa.vo.iris.visualizer.stil.StilPlotter;
 import uk.ac.starlink.ttools.plot2.Surface;
 import uk.ac.starlink.ttools.plot2.task.PlotDisplay;
 
 public class MouseCoordinateMotionListener extends StilPlotterMouseListener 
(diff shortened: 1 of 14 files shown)


eholum:
Added a unit test - We may need to think about a strategy for these kinds of tests (clicking, dragging, or other functions on the plotter?). Testing mouse movements seems like a tricky problem. For now I think everything is simple enough that testing the model is sufficient (when combined with a sanity check).

eholum:
Added additional functionality comments

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        Surface surface = display.getSurface();
+
+        Point p = evt.getPoint();
+        if (!surface.getPlotBounds().contains(p)) {
+            return;
+        }
+
+        double[] loc = surface.graphicsToData(p, null);
```
jbudynk: ah, that's how you get the x and y data from the Surface. Nice.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseListenerManager.java
```diff
@@ -0,0 +1,76 @@
...
+    private Set<StilPlotterMouseListener> listeners;
+    private PlotterView view;
+    
+    // Collection of all mouse listeners the plotter
+    MouseCoordinateMotionListener mouseCoordinateMotionListener;
+    PlotPointSelectionListener pointSelectionListener;
+    PlotPointSelectionDetailsListener pointDetailsListener;
+    
+    public MouseListenerManager() {
+        this.listeners = new LinkedHashSet<>();
```
jbudynk: why a linked hash set and not just a hashset or some other collection?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
eholum: ah yes, there's also stuff around there for zooming, etc... Some of it isn't in the generic surface interface, though.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPointSelectionDetailsListener.java
```diff
@@ -0,0 +1,133 @@
...
+import uk.ac.starlink.ttools.jel.ColumnIdentifier;
+import uk.ac.starlink.ttools.plot2.task.PlotDisplay;
+import uk.ac.starlink.ttools.plot2.task.PointSelectionEvent;
+import uk.ac.starlink.ttools.plot2.task.PointSelectionListener;
+
+/**
+ * Shows a tooltip with point name and details when selecting a point
+ * in the plotter.
+ *
+ * TODO: This functionality should ideally be done by a mouse hover function - but
```
jbudynk: so this TODO has nothing to do with the comment about asking for STIL's PlotDisplay class to switch  https://github.com/Starlink/starjava/blob/master/ttools/src/main/uk/ac/starlink/ttools/plot2/task/PlotDisplay.java#L405 from private to public, right? The PlotDisplay.createClickEvent() method still requires one to click on the plot for the nearest point,not to hover. Or could we extend this to listen for a hover event?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseListenerManager.java
eholum: Just here so that we always add them to the plot display in the same order. Doesn't really have any effect on anything now.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPointSelectionDetailsListener.java
eholum: It kind of does. We're not necessarily looking to switch that from public to private, but it would be nice to expose a method that gives the nearest point to any generic click coordinates. We could compute that manually ourselves by converting the mouse event to surface coordinates then checking to see which points are closest - but since there's already an implementation in Stil I don't really want to bother with it.

That being said if the segment data points are all always* sorted by spectral value we could do that pretty efficiently...

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPointSelectionDetailsListener.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+    
+    private void processEvent(PointSelectionEvent evt) throws Exception {
+        MetadataBrowserMainView mbView = this.getPlotterView().getMetadataBrowserView();
+        
+        long[] rows = evt.getClosestRows();
+        for (int i=0; i*2<rows.length; i++) {
+            if (rows[2*i] < 0) {
+                continue;
+            }
```
jbudynk: what's the point of this "if statement"?

eholum: I added an additional comment to my last commit (so refresh if you can). Basically the pointselectionevent will have a list of 'rows' for each layer in the plot display which correspond to the nearest point in that layer. If there is no such point, then the value is set to -1. 

The comments for both this and the other PointSelectionMouse listener have links to:

https://github.com/Starlink/starjava/blob/master/ttools/src/main/uk/ac/starlink/ttools/plot2/task/PointSelectionEvent.java#L60

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPointSelectionDetailsListener.java
```diff
@@ -0,0 +1,135 @@
...
+        } catch (Exception e) {
+            logger.warning(e.getMessage());
+        }
+    }
+    
+    private void processEvent(PointSelectionEvent evt) throws Exception {
+        MetadataBrowserMainView mbView = this.getPlotterView().getMetadataBrowserView();
+        
+        long[] rows = evt.getClosestRows();
+        for (int i=0; i*2<rows.length; i++) {
```
jbudynk: why is the for loop incremented by i*2?

eholum: Oh man, this is an interesting point. We add two layers for every star table (data and error bars), and so we end up with two layers in the PlotDisplay for every star table. Really this should increment by the number of layers in the segment layer.

So this works for now, but will break when we add options to turn off errors, or add additional layers. When we do that mapping from the point event list back into the list of star tables is going to be a bit trickier.

These point selection event listeners should have a common method that just converts a pointselectionevent to a startable and row number. I can add that for now, but how to handle more dynamic layers in the PlotDisplay/SegmentLayers will take some additional planning and though.

jbudynk:
Hi Erik,

I went through all the code except the test changes (which I'll look at later on this afternoon). It looks good. 

I noticed this PR removed the VisualizerComponentPreferences data member from the PlotterView, but still passes it on to the StilPlotter panel ([here](https://github.com/ChandraCXC/iris/pull/273/files#diff-d16d4d6a7d7ec5844bfa47dedeac896eL47) and [here](https://github.com/ChandraCXC/iris/pull/273/files#diff-d16d4d6a7d7ec5844bfa47dedeac896eL82)). I want to do some quick by-hand tests, but I'm glad to see it removed from the Plotter. We weren't doing anything with it in the PlotterView class anyhow.

I also noticed the north south east and west buttons were removed from the plotter. I'm not sure if I agree these should be removed in this PR. We should disable these buttons for now, but I still think they're a good feature to keep on the PlotterView. If a user doesn't have a mouse or trackpad, there would be no way for them to navigate the plot. Providing the zoom in/out and panning buttons lets those users navigate the plot.

jbudynk:
I tested the behavior by hand by importing a user and NED segment with the Test Builder. I can verify that selecting points in the plotter highlights them in the MB (ChandraCXC/iris-dev#29), and that the current cursor position is displayed in the Plotter X and Y boxes (ChandraCXC/iris-dev#15). Clicking on a point shows simple metadata (ChandraCXC/iris-dev#28).

There is some clunky behavior between displaying the simple metadata (ChandraCXC/iris-dev#28) and highlighting a point in the MB. If I click on a point to get it's metadata, the little "tooltip" pops up as expected. However, sometimes the tooltip does not disappear until I click somewhere else on the plot, or i move the mouse over and out of the tooltip. If the tooltip is still displayed, and I click on another point to highlight it in the MB, I have to click on that point twice.

My guess is something's going on with the `PlotPointSelectionListener` and `PlotPointSelectionDetailsListener` since both wait for a mouse click event.

jbudynk:
This wasn't a bug, but Erik agreed to update the behavior. The new commit have the tooltip popup underneath the cursor. When the cursor falls outside the range of the tooltip, the tooltip disappears. I'm able to easily click on multiple points quickly now. Thanks for the update!

PR #274, opened 2016-04-05
iris-dev#53 Show Column Metadata in Tooltip

eholum:
Closes https://github.com/ChandraCXC/iris-dev/issues/53

Added a custom JTableHeader to the IrisStarJTable class to support showing tooltips that look something like:

name: DataFluxStatErr
unit: Jy
utype: some.utype

Also fixed an issue with the units strings not being set on SegmentStarTable creation and updated the test to cover that case.

Supports permuting and resizing columns.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
+++ b/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
@@ -115,6 +115,10 @@ public class SegmentStarTable extends RandomStarTable {
         setSpecErrValues((double[]) getDataFromSegment(Utypes.SEG_DATA_SPECTRALAXIS_ACC_STATERR));
         setSpecErrValuesHi((double[]) getDataFromSegment(Utypes.SEG_DATA_SPECTRALAXIS_ACC_STATERRHIGH));
         setSpecErrValuesLo((double[]) getDataFromSegment(Utypes.SEG_DATA_SPECTRALAXIS_ACC_STATERRLOW));
+        
+        // Update column unit values
+        updateSpecColumnUnitStrings();
+        updateFluxColumnUnitStrings();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
```diff
@@ -163,26 +167,28 @@ public void setName(String name) {
...
+        
         // This may change Y values, so update them accordingly.
         setFluxUnits(fluxUnits);
-        
-        // Update column unit values
+    }
+    
+    private void updateSpecColumnUnitStrings() {
         for (int i=0; i<this.getColumnCount(); i++) {
             if (StringUtils.contains(getColumnInfo(i).getUtype(), "spec"))
```
jbudynk: don't all the SED-related utypes have "spec" in them? like, "Spectrum.Data. ..." Or are the ColumnInfo utypes different from the Spectral DM utypes?

jbudynk:
Other than the one comment about the utypes, this PR looks good. 

It'll be interesting to see how it works once we merge this and PR #275 (units widget). A test I'd like to do  is checking the units in the "Point Data" tab after changing them in the units widget.

iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
eholum: They're different. in the Column enum we manually set the Utypes for our "SegmentStarTables" .

See https://github.com/ChandraCXC/iris/pull/274/files#diff-1e7dd12e0894f768ddea60f2cef87aa3R375

Though we can change that to whatever. I just didn't want joining a segment star table to any other table to result in duplicate utypes in two different columns.

eholum:
Yeah that's a great call. This should always pull the units information out of the segment star table at runtime, so as long as all of that is being set it should be fine.

Definitely worth checking once everything is merged.

PR #275, opened 2016-04-11
Implement unit selection widget and add axes labels

jbudynk:
This PR addresses 

 - ChandraCXC/iris-dev#26  - Implement unit selection widget
 - ChandraCXC/iris-dev#62  - Add axes labels

Similar to Iris 2.1, clicking the "Units" button on the Plotter opens a frame in which the user can select the units they want to plot in. On clicking "Update," the currently selected SED's X and Y unit *preferences* update.

A `UnitsWidget` is used to display, select, and update a currently-selected SED's units. The units are updated via `UnitsWidget.update()` method, which calls `SedPreferences.setUnits()`, which updates the underlying `IrisStarTable` data.

The `SedPreferences.addSegment()` method was updated to allow segments StarTables already in the SED to be recreated with the new units.

**TODOs / TBDs**

- I would like the Widget to be able to support changing the units for a bulk of SEDs. This will be useful when we implement "global preferences" that can be applied to all the SEDs in the session (colors, shapes, sizes, units, etc.).
- Tests
    -  MB updates properly
    -  Plot updates as expected
- Both "flux" and "flux density" units are shown in the Manager. What shows in the YUnits list does not depend on whether "Flux" or "Flux Density" is chosen on the Plotter. Should we continue dividing these into separate groups, or list Flux and Flux Density units together?


iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/UnitsWidget.java
```diff
@@ -0,0 +1,247 @@
...
+import cfa.vo.iris.visualizer.preferences.SedPreferences;
+import cfa.vo.iris.visualizer.preferences.VisualizerChangeEvent;
+import cfa.vo.iris.visualizer.preferences.VisualizerCommand;
+import java.util.ArrayList;
+import java.util.List;
+import javax.swing.AbstractListModel;
+
+/**
+ *
+ * Unit selection widget. X-units are on the left split pane, Y-units are
```
eholum: Might be worth putting labels on the panel, though it should be pretty obvious to anyone looking at it...

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/UnitsWidget.java
```diff
@@ -0,0 +1,247 @@
...
+    private SedPreferences prefs;
+    private UnitsManager unitsManager;
+    
+    /**
+     * Creates new form UnitsWidget
+     * @param ws - workspace
+     * @param prefs - visualizer preferences
+     */
+    public UnitsWidget(IWorkspace ws, SedPreferences prefs) {
+        this.currentSed = (ExtSed) ws.getSedManager().getSelected();
```
eholum: I would attach the current sed through the preferences, rather than the workspace directly. Remember that this isn't thread-safe.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/UnitsWidget.java
```diff
@@ -0,0 +1,247 @@
...
+        this.prefs = prefs;
+        this.unitsManager = ws.getUnitsManager();
```
eholum: Is the the same units manager that's in the Default class? If so I would just grab the singleton directly from there. Then you could remove the workspace dependency.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedPreferences.java
```diff
@@ -112,7 +114,17 @@ void addSegment(Segment seg) {
         
         // If the segment is already in the map remake the star table
         if (segmentPreferences.containsKey(me)) {
+            
+            // TODO: this block might be slow. Is the table being updated twice?
             segmentPreferences.get(me).setInSource(convertSegment(seg));
+            
+            // reset the units
+            try {
+                segmentPreferences.get(me).setXUnits(xunits);
```
eholum: Didn't see this early, but it's probably worth replace all of the String class variables to the actually units type. Then we wouldn't need all of these try catches. And we could just pass the type around and classes could use it as needed -  so we have to convert from String to xunit all the time.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/UnitsManagerFrame.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Creates new form UnitsManagerFrame
+     * @param ws      workspace
+     * @param prefs   SED preferences
+     */
+    public UnitsManagerFrame(IWorkspace ws, VisualizerComponentPreferences prefs) {
+        this.ws = ws;
+        this.prefs = prefs;
+        initComponents();
+        this.currentSed = (ExtSed) ws.getSedManager().getSelected();
```
eholum: Same thing here, I would get the selected Sed from the preferences class.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/UnitsWidget.java
```diff
@@ -0,0 +1,247 @@
...
+    private javax.swing.JSplitPane jSplitPane1;
+    private javax.swing.JList xunits;
+    private javax.swing.JList yunits;
+    private org.jdesktop.beansbinding.BindingGroup bindingGroup;
+    // End of variables declaration//GEN-END:variables
+
+    public void updateUnits() {
+        // fire Visualizer event to update plot and MB
+        this.prefs.setUnits(xunit, yunit);
+        fire(currentSed, VisualizerCommand.RESET);
```
eholum: There's some weird behavior here, probably coming from the SedPreferences reset() method. Segments with an added suffix (e.g. 1) aren't keeping there same name and end up not being plotted when I apply a units change.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedPreferences.java
```diff
@@ -183,6 +200,22 @@ void setUnits(Segment seg, SegmentLayer layer) {
...
+     * @param yunit    the Y unit
+     */
+    public void setUnits(String xunit, String yunit) {
+        xunits = xunit;
+        yunits = yunit;
+        plotPreferences.setXlabel(xunit);
+        plotPreferences.setYlabel(yunit);
+        
+        // update the segment layers with the new units
+        refresh();
```
eholum: Just iterate over all of the segment layers in the map and set the units, rather than using refresh. Otherwise each segment will end up being re-serialized, which is super expensive especially for larger segments.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedPreferences.java
```diff
@@ -160,16 +173,20 @@ void removeSegment(Segment seg) {
     }
     
     /**
-     * Sets x and y units to the given SED if units are not already set.
+     * Sets x and y units of the given segment to the preferred units.
```
eholum: On closer inspection let's get rid of this method. It should be unnecessary and it is kind of useless.

Units shouldn't be set to the segments units, they should be sed to this segment layer's units.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedPreferences.java
jbudynk: At this point, I'm not going to refactor all the units from Strings to X or YUnit objects. There's a lot of classes that deal with units, and I'd prefer to just keep it safe for now. But it's a good idea to think about.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPreferences.java
```diff
@@ -111,7 +117,19 @@ public boolean getShowGrid() {
...
-        this.preferences.put(Y_LABEL, arg1);
+        String ylabel;
+        
+        try {
+            YUnits yconvert = new YUnits(arg1);
+            ylabel = yconvert.getLabel() + 
+                    " (" + SPVYUnit.getFromUnitString(arg1).getString() + ")";
+        } catch (SedException ex) {
+            // if unit is not defined
+            // TODO: set to default or unknown units?
```
eholum: At least log a warning so we know about it. Might even be worth throwing an exception, as this shouldn't* happen, right?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPreferences.java
```diff
@@ -120,7 +138,19 @@ public String getYlabel() {
...
-        this.preferences.put(X_LABEL, arg1);
+        String xlabel;
+        
+        try {
+            XUnits xconvert = new XUnits(arg1);
+            xlabel = xconvert.getLabel() + 
+                    " (" + XUnit.getFromUnitString(arg1).getString() + ")";
+        } catch (SedException ex) {
+            // if unit is not defined
+            // TODO: set to default or unknown units?
```
eholum: Ditto

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
```diff
@@ -92,6 +93,9 @@ public PlotterView(String title,
         plotter.setVisualizerPreferences(preferences);
         plotter.reset(null, true);
         
+        // units chooser frame
+        this.unitsManagerFrame = new UnitsManagerFrame(plotter.getSed(), this.preferences);
```
eholum: Just a though, what if the UnitsManagerFrame had a pointer to this plotter view instead of the preferences? Then in updateUnits could just be 

view.getPlotter().getSed() or whatever. And then you're guaranteed to be acting on the current plotted SED and don't need to have the extra VisualizerComponentListener.

jbudynk:
I added a logging message to the setXunit() and setYunit() methods in PlotPreferences, and I updated the way the units button actionPerformed() method opens the UnitsManagerFrame().

I am still working on the uispec4j tests for units selection.

jbudynk:
ah... another TODO:

 - handle magnitudes

If the Y axis is in magnitudes (ABMAG or STMAG), flip the Y range. From bottom to top, the numbers should decrease (lower magnitude = higher flux)

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
jbudynk: Thanks - makes sense. I made the changes in commit 19774b6

jbudynk:
I've added tests to check that the unit selector frame is working as expected.

One issue I noticed is the conversion between flux density errors and magnitude errors. The errors are as high as around +/- 30 mags, which is definitely wrong ;) I'll take a look at this bug before merging since it's related to units.

also: this is still pointing to "master" when it should be going to "sprint-8"

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPreferences.java
```diff
@@ -252,5 +311,20 @@ public PlotPreferences setLegendBorder(boolean arg1) {
...
+        String xlabel;
+        
+        try {
+            XUnits xconvert = new XUnits(arg1);
+            xlabel = xconvert.getLabel() + 
+                    "(" + XUnit.getFromUnitString(arg1).getString() + ")";
+        } catch (SedException ex) {
+            // if unit is not defined
+            // TODO: set to default or unknown units?
+            xlabel = "Wavelength (" + XUnit.ANGSTROM.getString() + ")";
```
eholum: whoops, log this, too.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedPreferences.java
```diff
@@ -183,6 +186,72 @@ void setUnits(Segment seg, SegmentLayer layer) {
...
+    /**
+     * Flips the Y-axis if yunit is a magnitude. A lower magnitude = brighter 
+     * source  higher on Y-axis)
+     * @param yunit 
+     */
+    private void flipYifMag(String yunit) {
+        try {
+            if (SPVYQuantity.MAGNITUDE.getPossibleUnits()
+                    .contains(SPVYUnit.getFromUnitString(yunit))) {
+                plotPreferences.setYflip(true);
```
eholum: Will this flip it back if we go from mag back to flux? Or does it automatically reset to false each time?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotPreferences.java
jbudynk: ah, nice catch - this function isn't even used at all - i'll remove it.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedPreferences.java
jbudynk: good catch - i've got a fix for this on my local machine that I'll push that flips the y-axis back.

jbudynk:
force pushed to add small fixes to erik's comments (flipping the y-axis back after switching and removing an unused method in PlotPreferences). Previous commit was b7cda8b.

eholum:
Looks good! Let's merge it (into sprint-8), and make sure any new work is captured in iris-dev for sprint-9.

jbudynk:
Closing this and reopening in PR #280 so it's based on sprint-8

PR #276, opened 2016-04-12
iris-dev#38 Export the Plotter as an Image

eholum:
Closes

https://github.com/ChandraCXC/iris-dev/issues/38

This PR maintains existing Iris2.1 functionality in terms of exporting an image plot (and does it in a cleaner manner IMO).

Initially I looked into what was in topcat for writing to an image file - and there is quite a bit of machinery to support different plot types, animations, etc that is beyond anything that we're looking for now. Since we're only dealing with simple, 2 dimensional plots that are contained entire in the StilPlotter, I just used a simple graphics object and Java's ImageIO to dump the StilPlotter into an image format of the user's choosing. If at some point in the future we decide to integrate more closely with Topcat's infrastructure, then it wouldn't be too much work to extend what's available in the 

https://github.com/Starlink/starjava/tree/6673d2b8eb4795a0a5d75300521e0ec4217a0df1/ttools/src/main/uk/ac/starlink/ttools/plot2/paper
and
https://github.com/Starlink/starjava/blob/6673d2b8eb4795a0a5d75300521e0ec4217a0df1/topcat/src/main/uk/ac/starlink/topcat/plot2/PlotExporter.java

To be able to pass a PlotDisplay and get back a buffered image io to write to a file.


jbudynk:
Looks good. I was able to open all the image formats. A few comments/questions:

High priority notes: I'd like to discuss or address these notes for this PR
- If there's a file of the same name, we should have a confirm message that says something like "File `<path/to/filename>` already exists. Do you wish to replace it?"
- If the file is saved successfully, should we have an info message that pops up saying "File saved to `</path/to/filename>`"?
- jpg and jpeg are the same format. I'd pick one of them and remove the other. I think ".jpg" is more common. But if you think both formats should be in the list, that's fine.

Maybe not for this PR (unless it's easy), but we'll want PostScript (.ps) and Encapsulated PostScript (.eps) formats, and maybe PDF too. 

Low priority notes: I'd say to tackle these tasks for this PR if it's not a hassle. Otherwise, no worries.
- I'm not sure I like where the ImageTypeChooser panel is. Could it be below the JFileChooser instead of to the left of it? If not, we could do what Iris does now: have a submenu for "Export to Image" where each submenu item is an image format.
- this item I don't care about too much, but I'll throw it out there (the other Iris "save" tools have the same problem).  
  If I type in a filename with the image extension included, the extension is printed twice. E.g., (1) if I choose "my-image.png" and select "png" format, the file is saved as "my-image.png.png". (2) If I choose "my-image.png" and select "jpg" format, the file is saved as "my-image.png.jpg". If the filename extension and image format are the same, then we shouldn't add the extra extension suffix to the filename. I.e., for (1) the filename would be "my-image.png" and for (2) would be "my-image.png.jpg".
- [low priority] adding an option to choose "Pixels per inch" and JPG quality to the image export

eholum:
- If there's a file of the same name, we should have a confirm message that says something like "File `<path/to/filename>` already exists. Do you wish to replace it?"

Added this.

- If the file is saved successfully, should we have an info message that pops up saying "File saved to `</path/to/filename>`"?

Added this.

- jpg and jpeg are the same format. I'd pick one of them and remove the other. I think ".jpg" is more common. But if you think both formats should be in the list, that's fine.

Removed jpeg. It was just there because it's in the standard image types supported by imageio.

- Maybe not for this PR (unless it's easy), but we'll want PostScript (.ps) and Encapsulated PostScript (.eps) formats, and maybe PDF too. 

It's not as simple, but it's also not that hard. Would probably take me an extra few hours to put it in. But I'm hoping that the basic images (which are the same as in Iris2.1) will get us there. If/When we're more tightly integrated with topcat all of this should hopefully go away (I added a question to our UDF). If it's worth it to have in the next release I'll spend the time to add it, otherwise I'd rather put it off till a later time.

- I'm not sure I like where the ImageTypeChooser panel is. Could it be below the JFileChooser instead of to the left of it? If not, we could do what Iris does now: have a submenu for "Export to Image" where each submenu item is an image format.

It's a pain in the but to move around the JFileChooser. Initially I had a panel with a JFileChooser in it but that doesn't look very good either, and it makes the behaviour of the window a little more annoying (like cancelling or not selecting, etc). For the file extension thing below, I added a little checkbox to add the extension (like in osx). It would also be easy enough to add a submenu as in Iris2.1. Let me know what you would prefer (I tend towards just having in the accessory).

- this item I don't care about too much, but I'll throw it out there (the other Iris "save" tools have the same problem).  
  If I type in a filename with the image extension included, the extension is printed twice. E.g., (1) if I choose "my-image.png" and select "png" format, the file is saved as "my-image.png.png". (2) If I choose "my-image.png" and select "jpg" format, the file is saved as "my-image.png.jpg". If the filename extension and image format are the same, then we shouldn't add the extra extension suffix to the filename. I.e., for (1) the filename would be "my-image.png" and for (2) would be "my-image.png.jpg".

Fixed.

- [low priority] adding an option to choose "Pixels per inch" and JPG quality to the image export

Yeah, not super easy to do - if we're just printing swing components we'd have to clone them to the size we wanted then print them. Or we break them up by file types as different imageIO formats support higher resolution, though I'm not sure it will print more than what's in the swing component anyway.

jbudynk:
Thanks for adding the "file saved" confirmation and the "file already exists" message. Looks good!

And I agree to your replies on moving the format selector and adding (E)PS and PDF formats.

Concerning the add/overwrite file extension, I'm not sure if this is the right approach either. I don't think I'd leave it as an option to the user. Instead, we should choose a behavior and force it on the user. Some options:

(1)  What I was suggesting is checking for the last ".ext" pattern in the input filename, and test to see if it matches the selected format. If it does, then ignore adding the extra extensions. If it doesn't match the selected format, add the selected format extension to the input filename

(2) OR, as what your Add/Overwrite option can do, we say if the filename contains an extension, and the extension matches one of the formats available, then save the file in that "overwritten" format instead of the one chosen in the drop-down menu.

(3)  OR, we don't add any file extension to the filename, and make the user type in the extension themselves. This is how Iris 2.1 works when saving SEDs.

Or, we stick with what you have now :) Maybe I'm overthinking this small part of the task. What do you think? What about @olaurino ?

eholum:
We'll go with option 1. Which is basically what we have now when the checkbox is marked (so I'll just remove that) - add it if it's not there or leave it if it is.

jbudynk:
yup, this looks great. thanks for the update.

jbudynk:
oh, just make sure you don't merge this into master. It should be towards sprint-8

eholum:
Ah, cheers, I didn't even notice (d'oh). I will NOT merge this into master.

eholum:
Closed in favor of https://github.com/ChandraCXC/iris/pull/277

PR #277, opened 2016-04-13
iris-dev#38 Export the Plotter as an Image

eholum:
Supersedes https://github.com/ChandraCXC/iris/pull/276. Reopen against sprint-8.

ChandraCXC/iris-dev#38

This PR maintains existing Iris2.1 functionality in terms of exporting an image plot (and does it in a cleaner manner IMO).

Initially I looked into what was in topcat for writing to an image file - and there is quite a bit of machinery to support different plot types, animations, etc that is beyond anything that we're looking for now. Since we're only dealing with simple, 2 dimensional plots that are contained entire in the StilPlotter, I just used a simple graphics object and Java's ImageIO to dump the StilPlotter into an image format of the user's choosing. If at some point in the future we decide to integrate more closely with Topcat's infrastructure, then it wouldn't be too much work to extend what's available in the

https://github.com/Starlink/starjava/tree/6673d2b8eb4795a0a5d75300521e0ec4217a0df1/ttools/src/main/uk/ac/starlink/ttools/plot2/paper
and
https://github.com/Starlink/starjava/blob/6673d2b8eb4795a0a5d75300521e0ec4217a0df1/topcat/src/main/uk/ac/starlink/topcat/plot2/PlotExporter.java

To be able to pass a PlotDisplay and get back a buffered image io to write to a file.


eholum:
Merged

PR #280, opened 2016-04-18
Implement unit selection widget and add axes labels

jbudynk:
(Originally PR #275) 

This PR addresses 

 - ChandraCXC/iris-dev#26  - Implement unit selection widget
 - ChandraCXC/iris-dev#62  - Add axes labels

Similar to Iris 2.1, clicking the "Units" button on the Plotter opens a frame in which the user can select the units they want to plot in. On clicking "Update," the currently selected SED's X and Y unit *preferences* update.

A `UnitsWidget` is used to display, select, and update a currently-selected SED's units. The units are updated via `UnitsWidget.update()` method, which calls `SedPreferences.setUnits()`, which updates the underlying `IrisStarTable` data.

The `SedPreferences.addSegment()` method was updated to allow segments StarTables already in the SED to be recreated with the new units.

**TODOs / TBDs**

- I would like the Widget to be able to support changing the units for a bulk of SEDs. This will be useful when we implement "global preferences" that can be applied to all the SEDs in the session (colors, shapes, sizes, units, etc.).
- Tests
    -  MB updates properly
- Both "flux" and "flux density" units are shown in the Manager. What shows in the YUnits list does not depend on whether "Flux" or "Flux Density" is chosen on the Plotter. Should we continue dividing these into separate groups, or list Flux and Flux Density units together?


jbudynk:
Rebased branch. Original tip was 7d38447.

PR #282, opened 2016-04-20
iris-dev#31 Applying Masks to StarTables

eholum:
Supersedes https://github.com/ChandraCXC/iris/pull/272 and https://github.com/ChandraCXC/iris/pull/269 and https://github.com/ChandraCXC/iris/pull/278

Closes

ChandraCXC/iris-dev#31
ChandraCXC/iris-dev#32

Additional minor changes were required in some of the ColumnInfoMatching logic in order to allow custom settings of default values for non-overlapping columns. And SegmentStarTables column logic has been improved and organized to support the new 'Filter Column' - which is displayed iff there is at least one filtered row in the selected tables.


eholum:
Rebased and reopened against sprint-9. Resolved behaviour conflicts between mouse point selection in the plotter and row selection in the metadata browser when there are masked points. Moved all of the stacking logic out of the metadata browser and into the IrisStarJTable - will make https://github.com/ChandraCXC/iris-dev/issues/68 a little easier.

eholum:
Added a method to the IrisStarJTable class to return an array of arrays for row selection - rather than having to map it back in later. This cleaned up the row selection logic to get rid of the weird mapping behavior in IrisStarTables - and is generally part of the larger refactoring picture. As it made this PR easier I put it here.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/masks/RowSubsetMask.java
```diff
@@ -0,0 +1,85 @@
...
+    public BitSet getMaskedRows(IrisStarTable table) {
+        // These filters only apply to a single star table.
+        if (this.table == table) {
+            return mask;
+        }
+        return new BitSet();
+    }
+    
+    @Override
+    public long cardinality() {
```
jbudynk: what does "cardinality" mean? What does it return?

jbudynk: oh, nevermind - I see it's in `BitSet`

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
```diff
@@ -137,8 +145,8 @@ public MouseListenerManager getMouseListenerManager() {
      * @return
      *  Collection of all the segment layers attached to the currently selected SED.
      */
-    public Collection<SegmentLayer> getSelectedLayers() {
-        SedPreferences p = getSelectedSedPreferences();
+    public Collection<SegmentModel> getSelectedLayers() {
+        SedModel p = getSedPreferences(getSelectedSed());
```
jbudynk: I like the name change from `SedPreferences` to `SedModel`. Should the other "sedPreferences" methods be changed to "sedModel" as well?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/IrisStarJTable.java
```diff
@@ -41,14 +49,51 @@
     // Use ColumnNames or Utypes in the column header fields
     private boolean utypeAsNames = false;
     
+    // Use the plotter data (segment) star tables or the metadata star tables
+    private boolean usePlotterDataTables;
```
jbudynk: is plotterTable the information in the "Data Tab," while segmentDataTable is the information in the "Metadata Tab" in the MB?

jbudynk: It might make sense to use the same buzz words in the comments as the code. For example, the variables, methods, and comments for the metadata don't match, and it makes it a little hard to read. Here, they're referred to as "metadata star tables," but below at line 91, it's `getSegmentDataTable()`. Also, the comment above implies the "plotter data" is the same as the "segment" data, but I think "segment data" means "metadata". Or am I misunderstanding something?

I think keeping the names similar in the comments and code will help with the readability. Let me know if you need a better explanation (here I am saying I find the code a little confusing, but I think my explanation above might be confusing too ;)).

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/IrisStarJTable.java
```diff
@@ -129,4 +205,52 @@ private String printColumnInfo(StarTableColumn column) {
...
+    public static class RowSelection {
+        
+        public final IrisStarTable[] selectedTables;
+        public final int[][] selectedRows;
+        public final int[] originalRows;
+        
+        private RowSelection(List<IrisStarTable> tables, int[] rows) {
+            
+            // Always sort the selection
+            Arrays.sort(rows);
```
jbudynk: why must the rows be sorted first? Do we not expect `rows` to be sorted already?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/IrisStarJTable.java
```diff
@@ -129,4 +205,52 @@ private String printColumnInfo(StarTableColumn column) {
...
+            int i = 0; // current selected row
+            
+            // Iterate through row selection and map rows to the appropriate star table
+            // row indexes
+            while (t < selectedTables.length) {
+                int end = (int) (startIndex + selectedTables[t].getBaseTable().getRowCount());
+                
+                // Get subarray the applies to this star table
+                int start = i;
+                while (i < rows.length && rows[i] < end) ++i;
```
jbudynk: is there no easier way to count `i`?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/IrisStarJTable.java
```diff
@@ -129,4 +205,52 @@ private String printColumnInfo(StarTableColumn column) {
...
+            int startIndex = 0; // start index of current star table
+            int t = 0; // current table
+            int i = 0; // current selected row
+            
+            // Iterate through row selection and map rows to the appropriate star table
+            // row indexes
+            while (t < selectedTables.length) {
+                int end = (int) (startIndex + selectedTables[t].getBaseTable().getRowCount());
+                
+                // Get subarray the applies to this star table
```
jbudynk:
```
// Get subarray *that* applies to this star table
```

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/ColumnMetadataStarTable.java
```diff
@@ -27,7 +28,8 @@
 import uk.ac.starlink.table.StarTable;
 
 /**
- * StarTable for producing a minimal metadata table for a list of StarTables.
+ * StarTable for producing a minimal metadata table for a list of StarTables. Will
+ * preserve an order on the startable if available in the ColumnInfoMatcher.
```
jbudynk: minor: did you mean
```
* preserve *the* order *of* the startable if available in the ColumnInfoMatcher
```
or no?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/IrisStarTable.java
```diff
@@ -16,18 +16,38 @@
...
+ * provides all necessary applications for viewing and manipulating data associated
+ * with a Segment. In particular,
+ * 
+ * 1) Is a container for the plotterStarTable - which maintains information associated
+ *  with the plotter (e.g. spectral and flux axis values).
+ * 2) Maintains a pointer to a the segment's metadata star table - which maintains all
+ *  metadata associated with the segment and each point in the segment.
+ * 3) Provides functionality for setting spectral and flux axis units for plotting data.
+ * 4) Allows filtering of data points for visualization on the plotter and for extraction
+ *  to the fitting tool.
```
jbudynk: nice! Comments here are super helpful.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/IrisStarTable.java
```diff
@@ -130,4 +171,163 @@ public static int getTableStartIndex(List<IrisStarTable> tables, IrisStarTable t
...
+    /**
+     * Remove the mask from rows on this StarTable.
+     * @param filter
+     */
+    public void clearMasks() {
+        mask = new RowSubsetMask(new int[0], this);
+        plotterTable.setMasked(mask.getMaskedRows(this));
+    }
+    
+    public BitSet getMasked() {
```
jbudynk: minor: add the javadoc since the methods around it have javadocs

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/IrisStarTable.java
```diff
@@ -130,4 +171,163 @@ public static int getTableStartIndex(List<IrisStarTable> tables, IrisStarTable t
...
+    
+    /**
+     * @return the filtered set of flux axis data values.
+     */
+    public double[] getFluxDataValues() {
+        return getFilteredValues(plotterTable.getFluxValues());
+    }
+    
+    /*
+     * Uses the BitSet masked to return a subset of data from the
```
jbudynk: typo:
```
* Uses the BitSet *mask* to return a subset of data from the
```

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/IrisStarTable.java
```diff
@@ -130,4 +171,163 @@ public static int getTableStartIndex(List<IrisStarTable> tables, IrisStarTable t
...
+        
+        int rows = (int) getRowCount();
+        double[] values = new double[rows];
+        
+        BitSet masked = mask.getMaskedRows(this);
+        int c = 0;
+        for (int i=0; i<(int) plotterTable.getRowCount(); i++) {
+            // Add only non-masked values.
+            if (!masked.get(i)) {
+                values[c++] = data[i];
```
jbudynk: would `data` ever be empty? if so, add a check for it

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/IrisStarTable.java
```diff
@@ -130,4 +171,163 @@ public static int getTableStartIndex(List<IrisStarTable> tables, IrisStarTable t
...
+            }
+        }
+        return values;
+    }
+    
+    /**
+     * We provide random access iff there are no filters applied to this
```
jbudynk: typo: `iff` should be `if`

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/SegmentColumnInfoMatcher.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public int compare(ColumnInfo c1, ColumnInfo c2) {
+        return getColumnForInfo(c1).compareTo(getColumnForInfo(c2));
+    }
+    
+    private Column getColumnForInfo(ColumnInfo c) {
+        try {
+            return Column.getColumn(c.getName());
+        } catch (IllegalArgumentException e) {
+            throw new RuntimeException("Cannot compare non-Column enum ColumnInfo objects", e);
```
jbudynk: Why is the error "Cannot compare non-Column enum ColumnInfo objects"? Shouldn't it be more like "Not a valid ColumnInfo object." or something like that? I see it's only used in the `compare()` method in this class, but it seems like an odd exception message if this method gets used for something other than comparisons.

But if you think this method won't be used elsewhere, then leave it as is

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/IrisStarJTableTest.java
```diff
@@ -70,4 +80,84 @@ public void testColumnHeaderTooltips() throws Exception {
...
+        // Should be two empty lists
+        RowSelection sel = table.getRowSelection();
+        assertEquals(2, sel.selectedRows.length);
+        assertEquals(0, sel.selectedRows[0].length);
+        assertEquals(0, sel.selectedRows[1].length);
+        
+        // Select all rows
+        table.addRowSelectionInterval(0, 5);
+        
+        // Should be two empty lists
```
jbudynk: update comment: should be two 3-element lists (all rows selected)

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/tables/MaskingTest.java
```diff
@@ -0,0 +1,259 @@
...
+        double[] y = new double[] {6,7,8,9,10};
+        
+        Segment seg = TestUtils.createSampleSegment(x, y);
+        
+        IrisStarTable test = adapter.convertSegment(seg);
+        
+        assertEquals(5, test.getRowCount());
+        seq = test.getRowSequence();
+        seq.next();
+        ArrayUtils.assertEquals(new Object[] {test.getName(), 1.0, 6.0, 6.0}, seq.getRow());
```
jbudynk: is this only looking at the metadata table? The "Data" tab would have 5 columns, right?

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/tables/MaskingTest.java
```diff
@@ -0,0 +1,259 @@
...
+        
+        // Move to end of StarTable
+        seq.next(); seq.next(); seq.next(); seq.next();
+        ArrayUtils.assertEquals(new Object[] {test.getName(), 5.0, 10.0, 10.0}, test.getRow(4));
+        
+        // Add a null filter
+        assertEquals(5, test.getRowCount());
```
jbudynk: this adds a null filter? or should the comment above be removed?

jbudynk:
Looks good! There were a couple of comments I made throughout the code for some simple issues/suggestions. I need to test the code by hand next, but for the most part, the source code looks fine.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
eholum: Ah, yeah probably, can I fold those changes into https://github.com/ChandraCXC/iris-dev/issues/68 as I'm trying to move all of those, anyway?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/IrisStarJTable.java
eholum: I totally know what you mean. I didn't think too much about it while working on it, but I'll go back through and make sure references are all lined up.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/IrisStarJTable.java
eholum: Not necessarily, mapping a row index from the view back into the model (from sorting - which is coming) might not preserve ordering of the selection.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/IrisStarJTable.java
eholum: I'll add a comment, we're not counting selected rows, but trying to figure out which rows in the selection apply to the current table from selectedTables.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/IrisStarJTable.java
eholum: fixed

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/ColumnMetadataStarTable.java
eholum: 'an order'

ColumnInfoMatchers support an ordering function if the caller wants one.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/IrisStarTable.java
eholum: Will be done.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/IrisStarTable.java
eholum: changed

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/IrisStarTable.java
eholum: It's only called on arrays specifically from the plotter table - they're non-null and guaranteed to be the same length.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/IrisStarTable.java
eholum: 'iff' is shorthand for 'if and only if' - but i'll expand it.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/SegmentColumnInfoMatcher.java
eholum: I'll change it

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/IrisStarJTableTest.java
eholum: changed

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/tables/MaskingTest.java
eholum: Yes but the underlying table only has 4 columns, the index column is attached to the jtable, not the startable.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/tables/MaskingTest.java
eholum: removed

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/SegmentColumnInfoMatcher.java
eholum: Actually I'll just remove it and let the illegalargumentexception from the enum propagate.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerComponentPreferences.java
jbudynk: Yeah, that makes sense

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/tables/ColumnMetadataStarTable.java
jbudynk: OK

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/stil/IrisStarJTable.java
jbudynk: OK

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/tables/MaskingTest.java
jbudynk: What do the values `1.0, 6.0, 6.0` represent? Is it x, y, and original y? If there were errors associated to the segment, would `seq.getRow()` return `{1.0, 6.0, <errValue>, 6.0}`?

eholum: Yeah, RowIterators iterate over each row in the star table, which for SegmentStarTables will always be made of the values in the SegmentColumn class (https://github.com/ChandraCXC/iris/pull/282/files#diff-6ae4f4a276800554797bd82ae4f7d409)

So in this case it's spectral, flux, original flux.

There are no error values in the generated test segments from TestUtils.

jbudynk:
Alright, I've tested the code by hand. Looks good for the most part! I noticed one bug, and I give a few suggested updates. We can figure out when we want to fix the bug / do the updates; they can be done here, or they can be added as new tasks - whatever works best.

**BUG**

"Remove Masks" doesn't work as expected when removing masks from multiple segments. Tested on Linux.

Different ways to see the issue:

- Add two SEDs from the test builder. Then add a NED SED of "3c066a". Open the MB, swtich to the Data tab, and mask one point in each segment (e.g., Index 21, 34, and 37). Then, highlight all the points with Ctrl+A, and click "Remove Masks". Depending on the row you clicked Ctrl+A on, one or two masked points will remain masked.
- Make sure all the masks are removed. Click on a row, then do Ctrl+A to select all the rows. Apply masks to all the rows. Then highlight all the rows again with Ctrl+A, and click "Remove Masks." Depending on the row you clicked on to select all the points, not all the masks are removed. Same thing happens if I use Shift+click to select all the points to remove the masks.
- If I select 3 masked points from 3 different segments, then click and highlight only those points, and click "Remove Masks," not all the masks are removed.

**SUGGESTIONS**

- an  easy one: I'd rename "Apply Masks" to "Mask Point" or "Mask Points," and rename "Remove Masks" to "Unmask". "Remove Masks" makes it sound like clicking that button will remove all the masks in my table without selecting the rows. Granted, this is my personal preference, so if you don't agree, let's talk about it.
- When I highlight some points to mask and click "Apply Masks", I expected to see the same points highlighted after the mask is applied. Is it possible to keep the selected points highlighted after the mask is applied? One benefit of this behavior:
    - Say I have some points already masked. I mask some more, but mistakenly pick the wrong points. If the same points that were just selected remain highlighted, I could just click "Remove Masks". 
- If I mask points from the Point Metadata tab, I don't see a "Mask" column appear; that column is only in the Data tab. It's a little unexpected. Is it easy to add a Masked column in the Point Metadata tab?
   - If the points remain highlighted after selection, in theory the same masked points should be highlighted in the Data tab, right? If it's not easy to add a Masked column to the Point Metadata tab, then keeping the points selected after masking them should be good enough for now. That way, the user can switch between the tabs and see their highlighted points were masked.

eholum:
Rebased this over sprint-9 to get Jamie's changes in here.

The bug is fixed by changes in https://github.com/ChandraCXC/iris/pull/293, which modifies the RowSelection further to support sorting.

Renamed the buttons based on Jamie's suggestions.

Added a new task for sorting/selection: https://github.com/ChandraCXC/iris-dev/issues/81

Added a new task for either showing the masked column or binding the selection models between the data and point metadata tabs: https://github.com/ChandraCXC/iris-dev/issues/82

jbudynk:
Thanks for making the changes and adding the new tasks. Looks good. OK to merge.

PR #283, opened 2016-04-20
iris-dev#58 Extract Selection to new SED

eholum:
Supersedes https://github.com/ChandraCXC/iris/pull/279

Closes https://github.com/ChandraCXC/iris-dev/issues/58

Allows users to use the 'Extract to New Sed' option in the file menu of the metadata browser to construct a new Sed with only the segments/points selected in the metadata browser.

The SegmentExtractor class basically just uses the clone method to create a copy of the original Segment, then copies in new points based on selection indexes from the Metadata Browser. I was initially worried about the extra overhead in double copying large sets of points, but it doesn't seem to affect performance in any significant way. If it comes to it we can always manually copy everything except for the ArrayOfPoint - it's just ugly.


eholum:
I have some changes in https://github.com/ChandraCXC/iris/pull/282 that would make this much simpler/cleaner, so maybe hold off reviewing this until I can rebase this over **https://github.com/ChandraCXC/iris/pull/282.

eholum:
Closed, currently in https://github.com/ChandraCXC/iris/pull/293 but will reopen for a deeper look next sprint.

PR #284, opened 2016-04-26
Add zoom button functionality

jbudynk:
This PR addresses the zoom in/out buttons previously disabled in PR #270. The zooming algorithm uses STILTS's [Axis::zoom() method](https://github.com/Starlink/starjava/blob/master/ttools/src/main/uk/ac/starlink/ttools/plot2/Axis.java#L486) to calculate the min and max range values. The tricky part of the algorithm is determining the central value of an axis, which depends on the axis spacing (log or linear).

Tests are in StillPlotterTest which test that zooming in once and zooming back out brings you back to the same view.

### DIFF
--- a/iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/StilPlotterTest.java
+++ b/iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/StilPlotterTest.java
@@ -176,4 +176,53 @@ public class StilPlotterTest {
         assertEquals(1, aspect3.getXMin(), .0001);
         assertEquals(1, aspect3.getYMin(), .0001);
     }
+    
+    // TODO: uncomment when we figure out why assertEquals() isn't working
+    //@Test
+    public void testZoom() throws Exception {
+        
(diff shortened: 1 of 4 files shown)


iris-visualizer/src/test/java/cfa/vo/iris/visualizer/stil/StilPlotterTest.java
```diff
@@ -176,4 +178,57 @@ public void testReset() throws Exception {
         assertEquals(1, aspect3.getXMin(), .0001);
         assertEquals(1, aspect3.getYMin(), .0001);
     }
-}
\ No newline at end of file
+    
+    @Test
+    public void testZoom() throws Exception {
+        
+        sed = ExtSed.read(TestData.class.getResource("3c273.vot").openStream(), SedFormat.VOT);
```
eholum: Just a thought, not an issue - this test would be simpler (and would run faster) to compute values if there were no segment - either just use the default axes from 0 to 10 or set the aspect to something easier to calculate.

eholum:
Looks good to me! Good choice on the 2/3 out vs in zoom. I think it looks great.

Merge it.

PR #287, opened 2016-04-27
Add usage instructions to the Plotter.

jbudynk:
Addresses ChandraCXC/iris-dev#9.

This PR adds a panel containing basic usage instructions for navigating about the Plotter. The instruction panel can be turned on/off through the Help menu of the Plotter.

To get the behavior I wanted when showing/hiding the usage instructions, I enclosed the Residuals buttons and usage instructions panels in a Swing `GridLayout`.

What I'd really like to do is add tooltips to the STILTS PlotDisplay when a
user hovers over the X and Y axes, but hover listeners in STILTS are private
methods as @eholum had mentioned in PR #273.


iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
```diff
@@ -389,33 +394,73 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
...
+
+        String navigationHelp1 = "<html>"+
+        "Scroll: zoom in and out<br/>"+
+        "Click and drag: pan the viewport</html>";
+        jLabel1.setFont(new java.awt.Font("DejaVu Sans", 0, 10)); // NOI18N
+        jLabel1.setText(navigationHelp1);
+
+        String navigationHelp2 = "<html>"+
+        "Shift + click + drag: zoom box<br/>"+
+        "Scroll on X or Y axis: expand/shrink along axis<br/>"+
```
eholum: Right clicking also does that weird zoom that I forget the name of...

eholum:
I like the messages. I'm a little wary about putting the jlabels right below the plot like that, though there is a ton of wasted space there. It'll either move or be between the plotter and the residuals. Could we put the instructions in a little pop-up window for now, since we don't have access to hovers? It could either open from the help option drop down, or we could put in a little help button (icon) somewhere in the frame. Thoughts?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
jbudynk: Ah right, thanks for catching that.

jbudynk:
I agree that I'm not the biggest fan of that panel showing up like that, which is why by default I have it hidden.

If I add it as a pop-up, I'd want it to look like an in-app user manual, but I don't know how to generate an in-app version of our user docs. I think this is something we have on our task list (though it should be a tier 2 or 3 task), or it's at least something we had talked about in the past.

I can make a pop-up for now that looks nice enough though :)

jbudynk:
I moved the plotter navigation docs to a JInternalFrame instead of leaving it under the view port. It has more details that may be useful for the iris-doc updates we'll eventually have to do.

The JInternalFrame has a JEditorPane that is populated by an HTML file added to the iris-visualizer resources directory.

There are no images in the user instructions. What would be best is if we could generate a JEditorPane from the HTML file generated by `mvn site` and have that (with its images) as the content help window. But for now, this HTML file should suffice. If anyone has suggestions/comments on the docs, let me know.

eholum:
This looks really good! I like the outside window - especially because it provides usage instructions for everything. Probably a good model for all of our components. Merge this.

PR #288, opened 2016-04-29
Save fit summary in human format

olaurino:
I will issue smaller PRs against the iris-dev-45 branch. That can't be a PR just yet because it doesn't have a diff with sprint-9.

# Description
This PR addresses part of chandracxc/iris-dev#45 by allowing users to save fit results in human readable text files. In particular, it fulfills req 2.2.b.1.

The only, small improvement I would like to make (pending review) is to add a timestamp when the file is saved, although we should really have a timestamp for fits.

This implementation has several improvements over the Iris 2.1 code. It includes data about the user models, as well as information regarding the optimizer and the cost function, which were missing in the old implementation.

Also, after an iteration with @jbudynk we decided to include all models in the output, without making a distinction between active and inactive components.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/fitting/FitConfiguration.java
+++ b/iris-common/src/main/java/cfa/vo/iris/fitting/FitConfiguration.java
@@ -36,6 +36,7 @@ public class FitConfiguration {
     private Stat stat;
     private Method method;
     private Confidence confidence;
+    private ConfidenceResults confResults;
     private Double rStat;
     private Integer nFev;
     private Double qVal;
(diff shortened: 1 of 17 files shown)


iris-common/src/main/java/cfa/vo/iris/fitting/FitConfiguration.java
```diff
@@ -36,6 +36,7 @@
     private Stat stat;
```
eholum: I'm not sure what the answer is for serializing this class - but there are plenty of tools out there that will handle java object i/o for it.

I'm generally a fan of JSON, but that's only because I'm most familiar with library support for it. So if it needs to be in a certain format then I'm sure there's already library support some where out there. I know that Jackson provides easy support for polymorphic types. Though off the top of my head I'm not sure what's available for saving/loading of proxied- though it wouldn't be too hard to write our own deserializers for loading the CompositeModel and Confidences.

I think we just pick a format, find a suitable library, then stick with it. Where else are we looking to use these saved configurations?

olaurino: The problem is not as much whether this is possible. It should work like a breeze with Jackson (which can also deserialize to interfaces). My question during the stand-up was more oriented whether we should try and reproduce the same format, thus also being forwards compatible, Iris 2.1 used.

I'll add a Json implementation, and then we'll discuss whether we should have some level of backward compatibility.

jbudynk:
easy issue: The majority of files are missing license headers. iris-visualizer/src/main/java/cfa/vo/iris/visualizer/FittingToolComponent.java and iris-visualizer/src/test/java/cfa/vo/iris/visualizer/FittingToolComponentTest.java should be updated to include 2016, too.

PR #292, opened 2016-05-04
Fitting Tool I/O

olaurino:
Resolve chandracxc/iris-dev#45.

# Description
This PR allows users to save fit results in human readable text files. In particular, it fulfills req 2.2.b.1.
It also enables users to save and read fit results in Json format, fulfilling req 2.2.b.2 (exept for the fact that the model is not evaluated, see task chandracxc/iris-dev#73).

We cannot read and write the older XML files at this point, as we discussed during the standup. We may add a task for a best-effort attempt to read in the old XML files.

The only, small improvement I would like to make (pending review) is to add a timestamp when the file is saved, although we should really have a timestamp for fits.

This human readable format has several improvements over the Iris 2.1 code. It includes data about the user models, as well as information regarding the optimizer and the cost function, which were missing in the old implementation.

Also, after an iteration with @jbudynk we decided to include all models in the output, without making a distinction between active and inactive components.


olaurino:
Rebasing. Original tip was at 83999ae.

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingMainView.java
```diff
@@ -487,4 +512,58 @@ public void mousePressed(MouseEvent e) {
...
+    }
+
+    private class SaveJsonAction extends AbstractAction {
+        public SaveJsonAction() {
+            super("Save Json...");
+        }
+
+        @Override
+        public void actionPerformed(ActionEvent actionEvent) {
+            int result = chooser.showOpenDialog(FittingMainView.this);
```
jbudynk: should be `chooser.showSaveDialog()` instead

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingMainView.java
```diff
@@ -487,4 +512,58 @@ public void mousePressed(MouseEvent e) {
...
+            if (result == JFileChooser.APPROVE_OPTION) {
+                try {
+                    controller.saveJson(new FileOutputStream(chooser.getSelectedFile()));
```
jbudynk: There be a catch for overwriting a file.

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingMainView.java
```diff
@@ -487,4 +512,58 @@ public void mousePressed(MouseEvent e) {
...
     }
+
+    private class SaveTextAction extends AbstractAction {
+        public SaveTextAction() {
+            super("Save Text...");
+        }
+
+        @Override
+        public void actionPerformed(ActionEvent actionEvent) {
+            int result = chooser.showOpenDialog(FittingMainView.this);
```
jbudynk: should be `chooser.showSaveDialog()`

iris/src/test/java/cfa/vo/iris/fitting/FittingFunctionalIT.java
```diff
@@ -70,6 +74,23 @@ public void testFittingThread() throws Exception {
...
+        saveText();
+    }
+
+    private void saveText() throws Exception {
+        File outputFile = tempFolder.newFile("output.fit");
+
+        WindowInterceptor
+                .init(fittingView.getMenuBar().getMenu("File").getSubMenu("Save Text...").triggerClick())
+                .process(FileChooserHandler.init().select(outputFile.getAbsolutePath()))
+                .run()
```
jbudynk: since this statement didn't catch the JFileChooser bug ([FittingMainView.java, line 523](https://github.com/ChandraCXC/iris/pull/292/files#diff-ff330886aa3ee19e24520f363ae5408eR523)), maybe we should use something else to test this functionality?

jbudynk:
Besides the issues in the comments I mentioned, the PR looks good! Once the comments are addressed -- preferably fixed, but at least discussed -- I'll give the OK to merge into master.

iris/src/test/java/cfa/vo/iris/fitting/FittingFunctionalIT.java
olaurino: Yes, it should be a saveDialog. However, FileChooser simply picks a file, so the functionality is fine and so is the test. I'll add an extra assertion for checking it is a save dialog.

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingMainView.java
olaurino: Definitely. However, we have a lot of ground to cover so if you agree I would add a task for fixing this and I would focus on other parts of the tool that are still missing or are significantly buggy first.

olaurino:
Commit f8fb816 should fix the `JFileChooser` save vs load dialog issue.

samp-factory/src/main/java/cfa/vo/sherpa/models/DefaultModel.java
```diff
@@ -88,7 +93,11 @@ public String getDescription() {
 
     @Override
     public void setName(String name) {
+        String[] tokens = name.split("\\.");
```
eholum: Potential NPE here

olaurino: Should be addressed by bbe830f

olaurino: Actually, I gave a closer look to StringUtils.split, and my fix probably doesn't cut it. I'll add a test and make sure I have a fix.

olaurino: @eholum Does 500f80b look reasonable?

eholum:
Yeah looks good. When this is green I say it's good to merge.

PR #294, opened 2016-05-06
Iris dev 56 filter expression validator

jbudynk:
Addresses ChandraCXC/iris-dev#56.

This PR adds a filter expression validator for the Metadata Browser. Currently, the `FilterExpressionValidator` constructor takes an `IrisStarJTable`, from which the column data is taken. [Javaluator](http://javaluator.sourceforge.net/en/home/) is used to parse and evaluate the expression. Filters can only consist of numeric columns. This is how the validator works:

- A few simple validity checks are made:
   - is the expression empty?
   - does the expression contain a column specifier (`$`)?
- `FilterExpressionValidator.findColumnSpecifiers()` finds the columns specified in the expression.
- Evaluate each row in the table. If the condition is not satisfied, the evaluator returns `Double.NaN` for that row.
   - Under the hood, the expression is tokenized. For example, the equation `$1 * 2 - 3 >= 10` becomes `("$1", "*", "2", "-", "3", ">=", "10")`. 
   - The evaluator finds the operators, determines the precedence of operations, and evaluates the expression in order of precedence. 
   -  `ComparisonDoubleEvaluator`, and extension of javaluator's `DoubleEvaluator` object, contains the logic for comparison/relational operators like `<, >=, ==`, etc. If the condition is not met, then the function returns `Double.NaN`. Default operators (`*, /, -, +, %, <functions>`, etc.) are handled by the parent class methods.
   - Various validity checks are made (even parentheses? valid column names?)
- The method returns the non-`NaN` row indices.

A `FilterExpressionException` is thrown if there is an issue with the filter expression. Default error messages are included in the class. The class is intended for use in integrating the filter expression validator with the Metadata Browser, so the end user gets an error message for any invalid expressions.

TODOs:

 - clean up code. There are currently a bunch of unused functions in `FilterExpressionValidator`, but these may be useful for...
 - Adding implementation for `AND, OR ANDNOT` logical operators
 - Add implementation for String-valued columns.


jbudynk:
Code now accepts logical operators NOT (`!`), AND (`&&`), and OR (`||`). However, logical operators cannot be grouped with parentheses yet. For example, this is OK:

```
2 * $1 - 5 >= 10 && $3 < 100 ! $2 <= 0 || $4 > 10
```

but this is not OK:

```
2 * $1 - 5 >= 10 && ($3 < 100 ! $2 <= 0) || $4 > 10
```

I would also like to let people use `AND, NOT, OR` instead of `&&, !, ||`, but as of now there are some Java regex issues I have to deal with.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/FilterDoubleExpressionValidator.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ *
+ * A class that validates an input filter expression for numeric data.
+ */
+public class FilterDoubleExpressionValidator {
+    
+    IrisStarJTable starJTable; // the stacked startable to filter
```
eholum: I would just have this depend on a StarTable (rather than the JTable) - if it's used with a starjtable the caller can specify the star table.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/ComparisonDoubleEvaluator.java
```diff
@@ -0,0 +1,139 @@
...
+     * If the expression does not contain relational operators, the method falls
+     * back on the DoubleEvaluator evaluate() method.
+     */
+    @Override
+    protected Double evaluate(Operator operator, Iterator<Double> operands, Object evaluationContext) {
+        Double left;  // left operand
+        Double right; // right operand
+
+        // for relational operators
+        if (operator == EQ) {
```
eholum: Rather than having a list of if/elses, what if you extended the operator class from above to something that supports operations on doubles?

class DoubleOperator extends Operator ...
    public abstract void evaluate(Double left, Double right)

You would still need a switch in a factory class for construction, but it makes this method dependent on an object, rather than cases. It's a pain to recast the operator, but I think it would clean this up a lot. Ditto the logical class. 

That being said, it's not entirely clear to me how these operators are parsed in javaluator, so if they can't be easily extended then it's not worth it.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/ComparisonDoubleEvaluator.java
```diff
@@ -0,0 +1,139 @@
...
+                return left;
+            }
+        } else if (operator == MINUS){
+            left = operands.next();
+            right = operands.next();
+            return left - right;
+            
+        } else {
+            // fall back on the parent method for other Double operators (*, -, +, etc.)
+            return super.evaluate(operator, operands, evaluationContext);
```
eholum: Ah okay, javaluator handles parsing the param into an operator and just passes it to this function then?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/FilterExpressionException.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package cfa.vo.iris.visualizer.metadata;
+
+/**
+ *
+ * Filter expression exception messages
+ */
+class FilterExpressionException extends Exception {
+
+    private static final String BAD_EXPRESSION = "Bad expression: ";
```
eholum: Most of these look like IllegalArgumentExceptions - which I would just throw instead of making a specific Exception class for this module. Though if you do have static string error messages it's easy enough to leave them in a class and reference them in the Exception construction.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/LogicalSetEvaluator.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+    
+    /**
+     * Get the union of set "a" with set "b". Changes set "a".
+     * @param a set to make unionize with b
+     * @param b set to compare and add to set a
+     * @return set "a" in union with set "b"
+     */
+    public static HashSet<Integer> union(HashSet a, HashSet b) {
```
eholum: Just use addAll?

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/FilterExpressionBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+        return set2list(hashSetRows);
+    }
+    
+    /**
+     * helper function: convert a set to a list
+     * @param set
+     * @return 
+     */
+    private List<Integer> set2list(HashSet set) {
+        List<Integer> list = new ArrayList<>();
```
eholum: ArrayList has a constructor that takes a collection that you can use instead.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/FilterExpressionBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ *
+ * Builds an expression tree of filters. Processing the filter expression tree 
+ * returns the table rows which follow the given filter.
+ */
+public class FilterExpressionBuilder {
+    
+    private LogicalSetEvaluator filterEvaluator;
+    
+    public FilterExpressionBuilder(IrisStarJTable table) {
```
eholum: Same deal, I would make everything dependent on a StarTable (or maybe even a list or iterator... StarTable is definitely good for now) - who knows where else this library might be used.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/FilterExpressionException.java
jbudynk: Thanks for taking a look at this one. I'm not sure what the best practice is: to make a new Exception class so we know exactly what exception to catch in the integration code (adding the filter validators to the MB), or if I should just let the javaluator exceptions get thrown without catching them first, or if there's a better way to handle these exceptions.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/LogicalSetEvaluator.java
jbudynk: Does addAll remove duplicates?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/ComparisonDoubleEvaluator.java
jbudynk: This is how the tutorial says one should extend their Evaluator classes. It's not pretty, but it works. I'm happy to use switch statements instead though.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/FilterExpressionException.java
eholum: Eh, I wouldn't let Javaluator specific Exceptions propagate, but in general I'm a fan of using built in exceptions whenever they satisfy what you're trying to communicate. If there were a specific action beyond just displaying an error message in the UI, then it would probably be worth making your own Exception. But IllegalArgumentExceptions, RuntimeExceptions, etc... let it pass up to the UI and it can handle them appropriately.

If we wanted something bigger, however, we should have specific service exceptions with error codes/messages and a top level handler that knew how to use them. That's in our backlog somewhere.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/ComparisonDoubleEvaluator.java
jbudynk: Yes, I'm not sure about the details, but the jist is their tokenizer adds operands and operators to the stack in order of precedence, then calls this evaluator function as it pops the right operators/operands off the stack. There are other `evaluate()` functions with different method calls (for Constant operands and Functions). I just use their default methods for these other `evaluate()`s.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/ComparisonDoubleEvaluator.java
eholum: Yeah all good. If that's the interface the service provides then this is cool, no difference to me on the switch vs if/else, though switch statements can be optimized by the compiler - so if you have a lot of else ifs, it's slightly more efficient.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/LogicalSetEvaluator.java
eholum: Depends on whether or not your set allows duplicates. In general a java set has the condition that it will never contain 2 elements e1 and e2 such that e1.equals(e2). I'm not sure what the operator.equals method look like, but the addAll method will do almost exactly what this code does.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/FilterDoubleExpressionValidator.java
```diff
@@ -0,0 +1,201 @@
...
+        // initial check for bad expressions
+        if (expression.isEmpty()) {
+            throw new FilterExpressionException(FilterExpressionException.EMPTY_EXPRESSION_MSG);
+        }
+        if (!expression.contains("$")) {
+            throw new FilterExpressionException(FilterExpressionException.DEFAULT_MSG);
+        }
+        
+        // get column specifiers
+        List<String> colSpecifiers = findColumnSpecifiers(expression);
```
eholum: I would have this function just return an array of column indexes - rather than the specifiers. Later on when we add more support we can do it there, rather than down on line 86/ with another for loop.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/FilterDoubleExpressionValidator.java
```diff
@@ -0,0 +1,201 @@
...
+                // throw an exception here:
+                try {
+                    colNumber = Integer.parseInt(colName);
+                } catch (NumberFormatException ex) {
+                    throw new FilterExpressionException(FilterExpressionException.NON_NUMERIC_COLUMN_NAME_MSG);
+                }
+                
+                try {
+                    if (!this.starJTable.getStarTable().getColumnInfo(colNumber).getContentClass().isAssignableFrom(String.class)) {
+                        variables.set("$"+String.valueOf(colNumber), (Double) this.starJTable.getStarTable().getCell(i, colNumber));
```
eholum: Actually, you might even break out the column index parsing into it's own class, the substitution of values doesn't need to be done here, but could be done on an expression/row by row basis. Make the column mapper class above, then have a function in it that takes a row in a star table and the expression and returns the evaluated expression. Thoughts?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/FilterDoubleExpressionValidator.java
```diff
@@ -0,0 +1,201 @@
...
+     * Process a simple expression and return a list of values
+     * @param expression
+     * @param col
+     * @return evaluated expression
+     * If no values in <tt>col</tt> match the condition, the function returns 
+     * null.
+     * @throws cfa.vo.iris.visualizer.metadata.FilterExpressionException if the 
+     *         expression is invalid.
+     */
+    public List<Double> processSingleExpression(String expression, double[] col) throws FilterExpressionException {
```
eholum: Where is this used?

jbudynk: ah! it's not haha. It's some left over code that isn't necessary anymore. Thanks for catching this.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/FilterExpressionBuilder.java
jbudynk: Ah, thanks - I've removed this function and just return `new ArrayList<>(hashSetRows)` in `FilterExpressionBuilder::process()` above.

olaurino:
Closing as this is obsoleted by #326.

PR #296, opened 2016-05-12
iris-dev#51,52,58 Sorting and Extraction in the Metadata Browser

eholum:
Supersedes https://github.com/ChandraCXC/iris/pull/283 and https://github.com/ChandraCXC/iris/pull/279

Closes

https://github.com/ChandraCXC/iris-dev/issues/58
https://github.com/ChandraCXC/iris-dev/issues/52
https://github.com/ChandraCXC/iris-dev/issues/51

Changes to support sorting on column values in the metadata browser, as well as extracting the current selection in the metadata browser to a new SED.

Builds on the RowSelection model in the IrisStarJTable class to give outside callers clean access to the underlying StarTables and rows, without having to deal with row sorting or ordering of the tables. The plotter data table in the metadata browser is sorted by spectral value by default. Order and selection is currently reset with the metadata browser, so applying masks will reset to default ordering and clear the selection. We have tasks in our backlog to address those issues at a later time.

SED extraction is handled by a separate class, and the new SED is asynchronously added to the the workspace by the visualizer preferences class. Existing infrastructure handles serializing/selected the new SED in the workspace.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
+++ b/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
@@ -178,6 +178,10 @@ public class SegmentStarTable extends RandomStarTable {
         }
     }
     
+    public Segment getSegment() {
+        return segment;
+    }
+    
     /**
      * Sets the spectral axis units for this star table, which updates all spectral
      * valued columns in the table. As this may also alter the flux axis units,
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/masks/RowSubsetMask.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/masks/RowSubsetMask.java
@@ -15,7 +15,6 @@
  */
 package cfa.vo.iris.visualizer.masks;
 
-import java.util.Arrays;
 import java.util.BitSet;
 
 import cfa.vo.iris.visualizer.stil.tables.IrisStarTable;
@@ -66,7 +65,7 @@ public class RowSubsetMask implements Mask {
     @Override
     public void clearMasks(int[] rows) {
         for (int i : rows) {
-            if (i > table.getSegmentMetadataTable().getRowCount()) continue;
+            if (i > table.getBaseTable().getRowCount()) continue;
             mask.clear(i);
         }
     }
@@ -78,7 +77,7 @@ public class RowSubsetMask implements Mask {
     @Override
     public void applyMasks(int[] rows) {
         for (int i : rows) {
-            if (i > table.getSegmentMetadataTable().getRowCount()) continue;
+            if (i > table.getBaseTable().getRowCount()) continue;
             mask.set(i);
         }
     }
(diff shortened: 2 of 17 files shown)


jbudynk:
The sorting looks good! I played around with masking and units and sorting, and it looks like the rows are sorted as expected.

I would suggest adding an "Extract" button to the bottom of the window with a tooltip "Extract selected points into a new SED" since Iris 2.1 has this.

However, something has broke keeping the same points highlighted in the Point Metadata and Data tabs. If I select points in the Data tab, and switch to the Metadata tab, I don't see the same points selected anymore. This is without any masks or sorting applied to either table.

eholum:
Sounds good. I'll add the button.

Row selection was deliberately removed between the two tables, as the sorting between them is not bound in anyway - see https://github.com/ChandraCXC/iris/pull/296/commits/088bb8fb4477cdcfc3db8438d847e9942a1daf77

If it's a critical feature then I can work on adding it back in, but it might be a good bit of effort. Before the selection models were just bound to each other, we could probably make a selection model that works on the underlying model rather than the view, but it's not immediately obvious to me if that's the right thing to do.

eholum:
Added an Extract button

jbudynk:
Thanks for adding the button.

2 things I noticed:

1. I think I'd change the created SED message to give the actual SED name, instead of calling each extracted SED "Filter SED". Or maybe something like "Added new extracted/filtered SED to workspace" (choose either "extracted" or "filtered"). But displaying the SED name is better I think; that way, people know what SED to look for in the Builder.

   In Iris 2.1, there isn't even a message. And since the MB now automatically updates with the new SED name, maybe we don't even need an info message?

2. The first extracted SED is added at the top of the Open SEDs list in the Builder. All the other SEDs added to the workspace are appended to the Open SEDs list.

jbudynk:
From an offline discussion with Erik:

1.  Adding the SED ID name to the info message. In commit febb598, adding an extracted SED to the Builder is asynchronous to the info message. Therefore, we can't know the SED ID at the time the info message is shown. Changeset 28b4bb6 removes the asynchronousity, which lets us know the name of the SED, but makes the operation slower, especially when extracting large numbers of points (e.g., high def spectra).

2.  Order of SEDs being added to Sed Builder. When creating/adding SEDs to the workspace before opening the Builder, SEDs are added to the SedManager, which stores SEDs in a TreeMap. The Builder stores them in a List.

olaurino:
couldn't you just create an empty SED to get the ID and then asynchronously add segments to it?

eholum:
Yes it absolutely can - and I adjusted it to do so. It also exposed a bug in renaming segments that also got addressed.

jbudynk:
I can confirm that the bug for renaming segments is fixed (ChandraCXC/iris-dev#69), and that the SED name is displayed in the error message correctly. Thanks for the fix!

Is this supposed to address ChandraCXC/iris-dev#85?

jbudynk:
Ah, spoke too soon. ChandraCXC/iris-dev#69 is not fixed, but this is a step in the right direction. The SED name is updated in the Visualizer and MB, but if I change the segment names, they still aren't updated in the legend or MB.

eholum:
Cool, thanks!

It does not address ChandraCXC/iris-dev#85, that's slightly more than a quick gimme, so saving it for a later time.

eholum:
Oh sorry, no ChandraCXC/iris-dev#69 is not fixed. It was an unrelated issue with refreshing the metadata browser and overlapping names in the data model not be correctly updated for existing segments that got fixed. Not ChandraCXC/iris-dev#69.

jbudynk:
Looks good. I give an OK to merge it

PR #297, opened 2016-05-25
iris-dev#68 Refactoring the Visualizer

eholum:
Closes https://github.com/ChandraCXC/iris-dev/issues/68

The majority of the changes are centred around moving the underlying data model for the metadata browser and the plotter into a common location - in particular, the VisualizerDataModel and the VisualizerDataStore. The VisualizerDataStore is responsible for maintaining all visualizer specific data w.r.t. the workspace sed manager. As such, the Sed change listeners have been moved out of the visualizer preferences class and into the VisualizerDataStore. Any time a class needs to look up a SedModel, they should use the DataStore. The VisualizerDataModel is the current runtime state of the visualizer - as such right now includes a List of selected Seds (currently attached to the Workspace's selected SED but that will change in https://github.com/ChandraCXC/iris-dev/issues/74. It also includes the currently selected segment models (the models highlighted in the metadata browser), and a few other details that can be used by any component to access or change the current state of the visualizer. Moving onto the shared datamodel allowed me to get rid of the whole visualizer change notifications mechanism, and the plotter and the metadata browser now use beans bindings to detect and redraw changes.

Other large changes include:
+ Override the equals method in ExtSed to include the id in equality checking. I'm not really a huge fan of the existing equals methods in Segment and Sed, and adding the equals check actually makes two empty seds with different ids different objects for use in HashMaps and propertyChangeNotifications.
+ Using a JTree in the metadata browser instead of a list. If the top level SED is highlighted all segments under the sed are selected, otherwise only the segments highlighted are selected.
+ Some changes in the Units frame - it's now created every time with the current datamodel, rather than just being hidden on close.
+ Added a layout manager to the plotter window.
+ Got rid of most of the resetlogic in the stil plotter - the reset method is not package protected, and should not be used by external callers. The fact that the Sed no longer switched in the middle of the method meant aspect storing logic had to change. We now store the aspect immediately before plotting only if we are re-drawing the same Sed.
+ Made the PlotPreferences values bindable, and bound the Grid settings in the plotter to them. There's more work to be done there when we have managed PlotPreferences windows, but for now the bindings seem to do the trick.
+ Slight changes in the logic for asynchronous serialization of segments into IrisStarTables - need to use the refresh method in the datamodel, which I'm not the biggest fan of.
+ I finally looked read through the caching mechanism in the StilPlotter, and can definitively say that we don't need it (yet).

All existing tests pass, and I've made adjustments to all tests to use the new DataStore and Model in their verifications.


eholum:
This includes the changes from https://github.com/ChandraCXC/iris/pull/296, so I would hold off reviewing this until that's merged.

olaurino:
What is the state of this PR? Is it reasonably safe to base work for ChandraCXC/iris-dev#77 on this branch and then rebase it?

eholum:
I'm not planning to do anything more major with this. That's all saved for https://github.com/ChandraCXC/iris-dev/issues/74, so I think you should be good to base off this. That being said, if you see anything major that might make residuals or model functions easier, let me know.

eholum:
Rebased onto master. Previously at d27960a.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/plotter/StilPlotterTest.java
```diff
@@ -231,4 +212,14 @@ public void testZoom() throws Exception {
         assertEquals(Math.log10(ymin), Math.log10(aspect1.getYMin()), 0.000001);
         
     }
+    
+    private StilPlotter setUpTests(ExtSed sed) throws Exception {
+        preferences = new VisualizerComponentPreferences(ws);
+        preferences.update(sed);
+        preferences.getDataModel().setSelectedSed(sed);
```
olaurino: I am surprised that you need to call two different methods from the same underlying class (`preference.update` delegates to the datastore, and so does `preferences.getDataModel` in order to set the SED. Shouldn't I be?

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/plotter/StilPlotterTest.java
```diff
@@ -231,4 +212,14 @@ public void testZoom() throws Exception {
...
+        
+        StilPlotter plot = new StilPlotter(preferences);
```
olaurino: Shouldn't a client be able to create a StilPlotter by simply using a SED as an argument? Maybe the resulting plotter is a static one that does not update if the SED is changed or a new SED is selected, but maybe through a different constructor or a specific method I should be able to instruct the plotter to react to such events by setting up the proper listeners in the underlying datastore. I would expect something similar in the metadata browser, and I would expect some tests to prove that in the non-static case changes to the underlying SED are picked up both by the plotter and the metadata browser, and that when a new SED is selected in the workspace, that event is handled too.

I am writing my comments as I go along (and while distracted by CIAO priorities coming and going), so I may be missing something.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/plotter/StilPlotterTest.java
eholum: Whoops, thanks for reminding me. I meant to remove all the operations for adding/removing references to the datastore from the preferences class - as maintaining a synchronous view as to what's in the sedmanager should be entirely up to the DataStore. Fixed that. The calls are to two different objects: The DataStore is the persistent store, so has no knowledge of what's in the plotter; whereas the DataModel is the underlying model that supports the current view. So this is first calling to put the SED into the map, then calling to select it in the view.

olaurino: But since you are calling everything from the preferences object you are constructing to initialize the plotter, maybe a different call (something like `preferences.setSed(sed)`) might be more convenient. In any case, as I commented below, I would expect one to be able to initialize a StilPlotter with a simple SED.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/plotter/StilPlotterTest.java
```diff
@@ -231,4 +212,14 @@ public void testZoom() throws Exception {
...
+        plot.resetPlot(false, false);
```
olaurino: Neither I would expect to need to reset the plot in order for the `StilPlotter` to work.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/preferences/SedPreferencesTest.java
```diff
@@ -36,65 +35,65 @@ public void testPreferences() throws Exception {
...
         
         // Ensure we get the right startables back
         assertEquals(3, layer1.getInSource().getRowCount());
         assertEquals(3, layer2.getInSource().getRowCount());
         
         // Remove a segment works
         sed.remove(seg1);
-        prefs.refresh();
-        assertEquals(1, prefs.getAllSegmentPreferences().size());
+        model.refresh();
```
olaurino: I realize this is a unit test, but shouldn't the refreshing happen automatically when events are fired? The sed was created with `managed=true` so events should be fired when a segment is added or removed.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/plotter/StilPlotterTest.java
eholum: The preferences object is the top level container for the data in the
visualizer - and provides access to whatever objects callers might need to
use to manipulate that data. The StilPlotter is not initialized by the
preferences, nor by the DataModel object, it just responds to and displays
changes in the model. Right now we're still very much in the Iris2.1
paradigm of being tied to the workspace, but going forward there's no
reason that has to be the case anymore. Ultimately the StilPlotter will
just respond what whatever DataModel you choose, and we could easily
construct multiple DataModels if users want to have multiple views open at
the same time.

On Thu, Jun 9, 2016 at 11:50 AM, Omar Laurino <notifications@github.com>
wrote:

> In
> iris-visualizer/src/test/java/cfa/vo/iris/visualizer/plotter/StilPlotterTest.java
> <https://github.com/ChandraCXC/iris/pull/297#discussion_r66467707>:
>
> > @@ -231,4 +212,14 @@ public void testZoom() throws Exception {
> >          assertEquals(Math.log10(ymin), Math.log10(aspect1.getYMin()), 0.000001);
> >
> >      }
> > +
> > +    private StilPlotter setUpTests(ExtSed sed) throws Exception {
> > +        preferences = new VisualizerComponentPreferences(ws);
> > +        preferences.update(sed);
> > +        preferences.getDataModel().setSelectedSed(sed);
>
> But since you are calling everything from the preferences object you are
> constructing to initialize the plotter, maybe a different call (something
> like preferences.setSed(sed)) might be more convenient. In any case, as I
> commented below, I would expect one to be able to initialize a StilPlotter
> with a simple SED.
>
> —
> <https://github.com/ChandraCXC/iris/pull/297/files/105aab45cc0c55029743d74faaeccfb5e3e5c9da#r66467707>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AHmNVcU1nqGmeoRXTAX_i3yuTEx3jpw1ks5qKDZPgaJpZM4ImmrL>
> .
>



-- 
Erik Holum

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/plotter/StilPlotterTest.java
eholum: It's unnecessary, leftover from pre-binding days. Removed.


On Thu, Jun 9, 2016 at 11:58 AM, Omar Laurino <notifications@github.com>
wrote:

> In
> iris-visualizer/src/test/java/cfa/vo/iris/visualizer/plotter/StilPlotterTest.java
> <https://github.com/ChandraCXC/iris/pull/297#discussion_r66469199>:
>
> > @@ -231,4 +212,14 @@ public void testZoom() throws Exception {
> >          assertEquals(Math.log10(ymin), Math.log10(aspect1.getYMin()), 0.000001);
> >
> >      }
> > +
> > +    private StilPlotter setUpTests(ExtSed sed) throws Exception {
> > +        preferences = new VisualizerComponentPreferences(ws);
> > +        preferences.update(sed);
> > +        preferences.getDataModel().setSelectedSed(sed);
> > +
> > +        StilPlotter plot = new StilPlotter(preferences);
> > +        plot.resetPlot(false, false);
>
> Neither I would expect to need to reset the plot in order for the
> StilPlotter to work.
>
> —
> <https://github.com/ChandraCXC/iris/pull/297/files/105aab45cc0c55029743d74faaeccfb5e3e5c9da#r66469199>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AHmNVWUzpDMMnBkEzgSC4ZVg9iDgLvncks5qKDgjgaJpZM4ImmrL>
> .
>



-- 
Erik Holum

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/preferences/SedPreferencesTest.java
eholum: It's a unit test for SedModel operations, not for the datastore, so the
event listeners aren't initialized for it. To make it more clear I'll set
managed to false.

On Thu, Jun 9, 2016 at 12:01 PM, Omar Laurino <notifications@github.com>
wrote:

> In
> iris-visualizer/src/test/java/cfa/vo/iris/visualizer/preferences/SedPreferencesTest.java
> <https://github.com/ChandraCXC/iris/pull/297#discussion_r66469859>:
>
> >
> >          // Ensure we get the right startables back
> >          assertEquals(3, layer1.getInSource().getRowCount());
> >          assertEquals(3, layer2.getInSource().getRowCount());
> >
> >          // Remove a segment works
> >          sed.remove(seg1);
> > -        prefs.refresh();
> > -        assertEquals(1, prefs.getAllSegmentPreferences().size());
> > +        model.refresh();
>
> I realize this is a unit test, but shouldn't the refreshing happen
> automatically when events are fired? The sed was created with managed=true
> so events should be fired when a segment is added or removed.
>
> —
> <https://github.com/ChandraCXC/iris/pull/297/files/105aab45cc0c55029743d74faaeccfb5e3e5c9da#r66469859>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AHmNVcVNEhh3K9fHkdmDx2nFmIaWbZ4Gks5qKDj0gaJpZM4ImmrL>
> .
>



-- 
Erik Holum

olaurino: So I guess the automatic refreshing is tested in the integration tests. I don't think we necessarily need a unit test, just checking :)

eholum: Yeah, the binding stuff is all caught by any visualizer test that uses
UISpec4j. This test is here because there's quite a bit of delicate logic
in naming, updating, etc for segments inside of an sed.

On Thu, Jun 9, 2016 at 12:19 PM, Omar Laurino <notifications@github.com>
wrote:

> In
> iris-visualizer/src/test/java/cfa/vo/iris/visualizer/preferences/SedPreferencesTest.java
> <https://github.com/ChandraCXC/iris/pull/297#discussion_r66473264>:
>
> >
> >          // Ensure we get the right startables back
> >          assertEquals(3, layer1.getInSource().getRowCount());
> >          assertEquals(3, layer2.getInSource().getRowCount());
> >
> >          // Remove a segment works
> >          sed.remove(seg1);
> > -        prefs.refresh();
> > -        assertEquals(1, prefs.getAllSegmentPreferences().size());
> > +        model.refresh();
>
> So I guess the automatic refreshing is tested in the integration tests. I
> don't think we necessarily need a unit test, just checking :)
>
> —
> <https://github.com/ChandraCXC/iris/pull/297/files/105aab45cc0c55029743d74faaeccfb5e3e5c9da#r66473264>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AHmNVRUzunt2bYOnzLfNePPCMdIcvfQVks5qKD0egaJpZM4ImmrL>
> .
>



-- 
Erik Holum

eholum:
Merged #297

PR #299, opened 2016-06-01
make asdc plugin a maven submodule

olaurino:
Resolve ChandraCXC/iris-dev#87 and make the ASDC plugin work again.

How far should we push this? Should I add proper tests for the plugin? To what extent?

### DIFF
--- a/iris-common/pom.xml
+++ b/iris-common/pom.xml
@@ -94,11 +94,6 @@
             <type>jar</type>
         </dependency>
         <dependency>
-            <groupId>org.jdesktop</groupId>
-            <artifactId>beansbinding</artifactId>
-            <version>1.2.1</version>
-        </dependency>
-        <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
             <version>1.4</version>
--- a/iris/pom.xml
+++ b/iris/pom.xml
@@ -56,11 +56,6 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>jsky</groupId>
-            <artifactId>jsky</artifactId>
-            <version>3.0</version>
-        </dependency>
-        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>iris-visualizer</artifactId>
             <version>${project.version}</version>
@@ -83,7 +78,7 @@
         </dependency>
         <dependency>
             <groupId>org.asdc.iris.plugin</groupId>
-            <artifactId>AsdcIrisPlugin</artifactId>
+            <artifactId>asdc-iris-plugin</artifactId>
             <version>1.1.7</version>
         </dependency>
     </dependencies>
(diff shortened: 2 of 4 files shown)


jbudynk:
I think we at least need a test to check that the ASDC component opens without errors.

If we wanted to add another test, it should be a simple integration test that checks that we can add a segments to the Builder. Given bug ChandraCXC/iris-dev#88, we should make sure there aren't too many segments selected from the ASDC query.

olaurino:
@jbudynk the trick with the "simple" integration test is that it must rely on the asdc service, which may make the tests flaky, because of potential connectivity issues at test time. I'll try adding one nevertheless, let's hope it doesn't backfire.

eholum:
Can you stub out the asdc plugin with preloaded data? Either that or just
add a simple test to make sure the plugin is open-able. We shouldn't spend
time on getting a good integration test in now, just something that makes
sure it isn't totally dead.

On Wed, Jun 1, 2016 at 4:27 PM, Omar Laurino <notifications@github.com>
wrote:

> @jbudynk <https://github.com/jbudynk> the trick with the "simple"
> integration test is that it must rely on the asdc service, which may make
> the tests flaky, because of potential connectivity issues at test time.
> I'll try adding one nevertheless, let's hope it doesn't backfire.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/ChandraCXC/iris/pull/299#issuecomment-223114498>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AHmNVb8i06iAQxJHRC7k6cbxfGpGVmDeks5qHetPgaJpZM4Ir06n>
> .
>



-- 
Erik Holum

olaurino:
@eholum I agree, plus I was looking into it and a "simple" test without too many segments requires fiddling with a custom jtree with a tristate checkbox cell renderer. It is not immediately obvious how to stub the component to return pre-loaded data, and in any case that would not check we can actually use the plugin: if their servlet starts serving bogus data we wouldn't catch that. The plugin is not going to change any time soon, so that test would not mean much.

I'll push the most basic test I got running, and we can include more tests (at very low priority) later.

jbudynk:
That sounds fine to me. As long as we can catch that the plugin at least opens, that's a good start. I agree with @olaurino that we'll eventually want a test that does a simple query with the actual ASDC service. If they change their service in any way that breaks our component, we'd want to know about it. Does it need to be done in this PR? Not sure -- depends on the effort involved. If it's a decent amount of effort, I'd be OK with putting it in a separate task.

olaurino:
Interactive tests show that the plugin is currently unusable, most likely because of issues on our side. Adding tests now does not seem to make sense to me, but this is something we should address, and it looks like it should have a higher priority than the one we discussed this morning regarding seds with many segments.

olaurino:
Rebasing, original tip was 19f95f8.

eholum:
Confirmed it opens on linux and macs - assuming this next build passes, go ahead and merge.

PR #300, opened 2016-06-01
iris-dev#100 Fix Tooltip in Metadata Browser

eholum:
Resolves https://github.com/ChandraCXC/iris-dev/issues/100

This is unrelated to https://github.com/ChandraCXC/iris-dev/issues/101.

Basically just had a stray empty string in the tooltip setting for one of the metadata browser panels. Also added a check to make sure we don't ask for the -1 indexed column in the startable - which can happen if the header doesn't fill the entire frame.


olaurino:
I can confirm this is fixed. As long as travis is happy with it, we can probably merge this.

eholum:
Merged #300

PR #301, opened 2016-06-10
Fix issue with AB magnitudes not being converted correctly.

jbudynk:
Addresses ChandraCXC/iris-dev#72

This PR fixes an issue where conversions between magnitudes and flux/flux densities were not done correctly.

First, the wrong units conversion method was being used in [`SegmentStarTable::setFluxUnits()`](https://github.com/ChandraCXC/iris/blob/master/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java#L221). Before, it was using `UnitsManager::convertY()`. Now, it uses `UnitsManager::convertErrors()` for the flux errors which has a special cases for converting errors to/from magnitude units (in the Specview units code).

Next, Specview's `YUnits` conversion method for AB magnitudes wasn't perfect. This has been fixed in line [862](https://github.com/ChandraCXC/iris/blob/iris-dev-72-magnitudes-conversion/iris-common/src/main/java/cfa/vo/iris/units/spv/YUnits.java#L862). This is what was causing the flux values to change when the user converts back and forth between ABMAG and another unit multiple times.

Converting errors between magnitude and flux/flux density units, however, is more tricky. Even doing calculations by hand, I haven't figured out how to convert the errors back and forth between mags and flux. So as a solution, this PR stores *all* the original flux values and errors, along with the original units, and always starts from here when converting between units. I am not happy with this solution, as this takes up more space. However, now we could have a method to reset the SED back to it's original state without any issues.

TODOs:
   - make sure the OriginalFlux column in the MB shows the original units.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
+++ b/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
@@ -57,6 +57,7 @@ public class SegmentStarTable extends RandomStarTable {
     private Segment segment;
     private XUnit specUnits;
     private YUnit fluxUnits;
+    private YUnit originalFluxUnits;
     private final long rows;
 
     // Columns for name and filtered state
@@ -70,6 +71,9 @@ public class SegmentStarTable extends RandomStarTable {
     private double[] specValues;
     private double[] fluxValues;
     private double[] originalFluxValues;
+    private double[] originalFluxErrValues;
+    private double[] originalFluxErrValuesHi;
+    private double[] originalFluxErrValuesLo;
     private double[] specErrValues;
     private double[] specErrValuesLo;
     private double[] specErrValuesHi;
@@ -88,6 +92,7 @@ public class SegmentStarTable extends RandomStarTable {
         this.segment = segment;
         this.specUnits = units.newXUnits(segment.getSpectralAxisUnits());
         this.fluxUnits = units.newYUnits(segment.getFluxAxisUnits());
+        this.originalFluxUnits = units.newYUnits(segment.getFluxAxisUnits());
         this.columns = new TreeSet<SegmentColumn>();
         
         // Name column will always be first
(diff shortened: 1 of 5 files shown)


iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
```diff
@@ -116,6 +121,9 @@ public SegmentStarTable(Segment segment, String id)
         setSpecValues(segment.getSpectralAxisValues());
         setFluxValues(segment.getFluxAxisValues());
         setOriginalFluxValues(segment.getFluxAxisValues()); // Copies data so these aren't the same
+        setOriginalFluxErrValues((double[]) getDataFromSegment(Utypes.SEG_DATA_FLUXAXIS_ACC_STATERR));
```
eholum: I would re-use the values from below - SedLib is REALLY slow at these methods - remember the slow loading issue when the 300k spectra is loaded first? It's because of these methods... which we need to get rid of somehow. 

After the setFlux/SpecErr values below, I would just call

setOriginalFluxErrValues(getFluxErrValues())

Rather than using sedlib twice.

jbudynk:
I talked with Ian about what I was doing wrong with the magnitude error conversions. There is a negative sign that I was missing. Turns out that the specview code also neglects this negative sign. However, I can explain why I think the specview code is missing this sign:

The conversion from flux density (say, erg/s/cm2/Hz) to AB magnitudes is

```
m_ab = -2.5 * log10(f_nu) - 48.6
m_ab_err = -2.5 * log10(1 + f_nu_err/f_nu)
```

The magnitude errors will be negative for some cases. Of course we don't want to report the errors back to the user as being negative, as that doesn't make much sense. So, the errors should be shown as positive numbers.

Here is the equation for converting back from magnitudes to flux density:

```
f_nu = 10**(-0.4 (m_ab + 48.6) )
f_nu_err = 10**(-0.4 (m_ab + m_ab_err + 48.6)) - f_nu
```

If the magnitude error sign isn't considered, you can see that you'll get back an erroneous flux density error. I was forgetting this sign in my calculations by hand, and so is specview. Neglecting the error sign, I get the same erroneous uncertainty values as specview.

So, we can stick with what the PR currently has - always go back to the original flux and error values - or, we can change the specview code. I've been trying to see how we can keep track of the error sign, but it seems cumbersome. For now, I'll leave the PR as-is, plus Erik's suggestion.

eholum:
Looks good. Man, when originally implementing the segment star table I did not realize how many columns there were going to be. Might be worth looking at sometime in the future.

Merge this!

jbudynk:
Rebased onto master. Original tip was 538de2f

PR #303, opened 2016-06-16
iris-dev#74 Support Co-Plotting

eholum:
Resolves: https://github.com/ChandraCXC/iris-dev/issues/74

PR for supporting coplotting in the Iris Visualizer.

**Major Changes**
* The logic for the relationship bewteen the workspace and the plotter has been implemented to replicate what's currently in Iris2.1 - so switching selected SEDs will still change what's in the plotter.
* Primary support for the change comes from removing the idea of a 'selectedSed' in the VisualizerDataModel - replaced instead with a list of 'selectedSeds' that are passed through the Coplotting management window. All components that depend on the dataModel rely on that interface.
* Generalized the SegmentModel to that of a LayerModel - which now takes a generic startable instead of an IrisStarTable and is the primary representation of a 'layer' in the StilPlotter. Still supports plotting both errors and marks - though easily extended/adjusted for lines and other layer types. 
* PlotPreferences have been moved out of the SedModel and into the VisualizerComponentPreferences class. Preferences are stored in a weak referenced, identity hashmap keyed off either Lists of SEDs or off of SEDs themselves. Lists of SEDs are ephermeral and will only exist while coplotting, so after exiting a coplot they will disappear.
* Units logic as it relates to plotting has been cleaned up further, most of the actual logic has been moved to the VisualizerDataModel, which also handles converting underlying StarTables to the preferenced units before plotting, if available.
* When coplotting, the entire SED is put into a StackedStarTable and plotted as a single layer.

**Other Changes**
* Plot Titles/Window names are simply a list of the ids of every SED in the datamodel.
* Logic for how the dropdown windows in the MetdataBrowser are preset has changed slightly. If there's only one sed in the list it will be fully expanded, otherwise all SEDs are not expanded. The list is fully selected by default.
* Additional logic was needed in the StilPlotterMouseSelection listener to map the layers back into the correct star table and row index. Additional unit tests have been added as that logic is somewhat tricky.
* I added a function in ExtSed to get an unmodifyable List of segments. There are so many places that requiring iterating over the list it seems silly to not provide an interface to access it directly.
* There is some logic for 'freezing' the workspace that was removed by https://github.com/ChandraCXC/iris/commit/d1f11a4d668097b601d2f55f525f76487f42dd22. For now it requires more work to hashout the details, so I removed it but left the commit and a TODO as a reference to a starting point.


iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/MouseCoordinateMotionListener.java
```diff
@@ -38,7 +39,7 @@ public void setPlotterView(PlotterView plotterView) {
     }
 
     @Override
-    public void activate(PlotDisplay<?,?> display) {
+    public void activate(PlotDisplay<?,?> display, VisualizerDataModel dataModel) {
```
jbudynk: it doesn't look like the DataModel object is used

eholum: It's not in this class, but it's available if other objects that define that interface.

jbudynk: Ah, ok. thanks

jbudynk:
The Plotter window width is a little too narrow on the OS X build (I haven't check Linux yet). I can't see the "Units" button without expanding the window size.

eholum:
Okay, messed around quite a bit with the layout of the plotter to try to get them to play nicely on both macs and linux, to my eyes it looks okay, but let me know if there are any other issues.

eholum:
Rebased over master, previous tip at d9b4fa0

olaurino:
I tested it interactively and I exercised some of the API in #305, it looks good!

Since we already based three PRs off of this one, I would rather merge it, and if we find something wrong we can fix it as a bug, but if we changed something while this is still not merged, it would be a mess :)

eholum:
Yeah, I very much agree with you. If anyone comes across anything that they don't like - feel free to open a task for it, or just incorporate it into the changes you're working on if it's small enough.

eholum:
Merged #303

PR #304, opened 2016-06-23
Plot fitted model

jbudynk:
Adds the capability of plotting a model on top of an SED.

This PR provides a very simple visualization of a plotted model. It assumes that the evaluated model is a SegmentStarTable. The method is called with `StilPlotter::plot_model(SegmentStarTable model)`.

A `FunctionModel` is the evaluated model's (or function's) "model", which contains the STILTS preferences for plotting a StarTable as a line. It is similar to the `SegmentModel` object. It's list of preferences is shorter: you can control the line thickness, color, and pattern. For now, the defaults are the same as STILTS' defaults: a solid, red line of thickness 1 (thin like Iris 2.1's line).

To take a look at the model on the visualizer, I've added a dummy "Plot Model" functionality to the PlotterView frame. It plots a FunctionModel version of the currently selected SED. I expect to remove/replace this functionality when we have fitting notifications defined.

Outstanding questions:

 - What should happen if a user overplots a relatively large model on a short SED? I.e., if the model's x range and/or y range >> SED's x and y range. Should the plot aspect increase to include the entirety of the model, or should it stay focused on the SED's data range? No matter what, users could still zoom in/out of the plot as usual. This question would determine how the `resetPlot()` method works when plotting functions.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
@@ -256,7 +256,6 @@ public class PlotterView extends JInternalFrame {
         mntmAutoFixed = new javax.swing.JCheckBoxMenuItem();
         mntmGridOnOff = new javax.swing.JCheckBoxMenuItem();
         mntmCoplot = new javax.swing.JMenuItem();
-        mntmShowModel = new javax.swing.JCheckBoxMenuItem();
         mnHelp = new javax.swing.JMenu();
         mntmPlotterNavigationHelp = new javax.swing.JMenuItem();
 
@@ -605,14 +604,6 @@ public class PlotterView extends JInternalFrame {
         });
         mnView.add(mntmCoplot);
 
-        mntmShowModel.setText("Show model");
-        mntmShowModel.addActionListener(new java.awt.event.ActionListener() {
-            public void actionPerformed(java.awt.event.ActionEvent evt) {
-                mntmShowModelActionPerformed(evt);
-            }
-        });
-        mnView.add(mntmShowModel);
-
         menuBar.add(mnView);
 
         mnHelp.setText("Help");
@@ -695,29 +686,6 @@ public class PlotterView extends JInternalFrame {
         plotter.dataPan(SwingConstants.WEST);
     }//GEN-LAST:event_rightActionPerformed
 
-    private void mntmShowModelActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_mntmShowModelActionPerformed
-        
-        // Plot a model
-        // TODO: This is a dummy model; it's just a line plot of the first segment
-        // in the current SED. We need to replace this with the fitted model
-        ExtSed sed = plotter.getDataModel().getSelectedSeds().get(0);
-        SedModel model = plotter.getDataModel().getSedModel(sed);
-        
-        if (mntmShowModel.isSelected()) {
-            try {
-                model.getDataTables().get(0).getPlotterDataTable().setModelValues(sed.getSegment(0).getFluxAxisValues());
-                FunctionModel functionModel = new FunctionModel(model);
-                plotter.getDataModel().getSedModel(plotter.getDataModel().getSelectedSeds().get(0)).setFunctionModel(functionModel);
-                plotter.resetPlot(false, false);
-            } catch (SedNoDataException ex) {
-                Logger.getLogger(PlotterView.class.getName()).log(Level.SEVERE, null, ex);
-            }
-        } else {
-            plotter.getDataModel().getSedModel(plotter.getDataModel().getSelectedSeds().get(0)).setFunctionModel(null);
-            plotter.resetPlot(false, false);
-        }
-    }//GEN-LAST:event_mntmShowModelActionPerformed
-
     // Variables declaration - do not modify//GEN-BEGIN:variables
     private javax.swing.JPanel bottomButtonsPanel;
     private javax.swing.JButton btnReset;
@@ -744,7 +712,6 @@ public class PlotterView extends JInternalFrame {
     private javax.swing.JMenuItem mntmPlotterNavigationHelp;
     private javax.swing.JMenuItem mntmProperties;
     private javax.swing.JMenuItem mntmSave;
-    private javax.swing.JCheckBoxMenuItem mntmShowModel;
     private javax.swing.JMenuItem mntmSomething;
     private javax.swing.JRadioButtonMenuItem mntmXlog;
     private javax.swing.JRadioButtonMenuItem mntmYlog;
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
@@ -602,7 +602,6 @@ public class PlotterView extends JInternalFrame {
         });
         mnView.add(mntmCoplot);
 
-        jCheckBoxMenuItem1.setSelected(true);
         jCheckBoxMenuItem1.setText("jCheckBoxMenuItem1");
         jCheckBoxMenuItem1.addActionListener(new java.awt.event.ActionListener() {
             public void actionPerformed(java.awt.event.ActionEvent evt) {
@@ -700,13 +699,17 @@ public class PlotterView extends JInternalFrame {
             // TODO: right now, it just overplots a flattened-version of the SED to
             // play around with. This should be updated to show the currently
             // selected SED's model, if it exists.
-            
-            ExtSed sed = preferences.getDataModel().getSelectedSeds().get(0);
-            String xunits = preferences.getDataModel().getSedModel(sed).getXUnits();
-            String yunits = preferences.getDataModel().getSedModel(sed).getYUnits();
 
-            SegmentStarTable model = new SegmentStarTable(ExtSed.flatten(sed, xunits, yunits).getSegment(0), title);
-            plotter.plot_model(model);
+            if (jCheckBoxMenuItem1.isSelected()) {
+                ExtSed sed = preferences.getDataModel().getSelectedSeds().get(0);
+                String xunits = preferences.getDataModel().getSedModel(sed).getXUnits();
+                String yunits = preferences.getDataModel().getSedModel(sed).getYUnits();
+
+                SegmentStarTable model = new SegmentStarTable(ExtSed.flatten(sed, xunits, yunits).getSegment(0), title);
+                plotter.plot_model(model);
+            } else {
+                plotter.remove_model();
+            }
         } catch (SedException | UnitsException ex) {
             Logger.getLogger(PlotterView.class.getName()).log(Level.SEVERE, null, ex);
         }
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
@@ -706,7 +706,7 @@ public class PlotterView extends JInternalFrame {
                 String yunits = preferences.getDataModel().getSedModel(sed).getYUnits();
 
                 SegmentStarTable model = new SegmentStarTable(ExtSed.flatten(sed, xunits, yunits).getSegment(0), title);
-                plotter.plot_model(model);
+                plotter.plotModel(model);
             } else {
                 plotter.remove_model();
             }
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
@@ -18,14 +18,19 @@ package cfa.vo.iris.visualizer.plotter;
 import cfa.vo.iris.IWorkspace;
 import cfa.vo.iris.IrisApplication;
 import cfa.vo.iris.gui.GUIUtils;
+import cfa.vo.iris.sed.ExtSed;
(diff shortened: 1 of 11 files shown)


iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
```diff
@@ -570,12 +584,36 @@ private void metadataButtonActionPerformed(java.awt.event.ActionEvent evt) {//GE
         }
     }//GEN-LAST:event_metadataButtonActionPerformed
 
+    private void jCheckBoxMenuItem1ActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_jCheckBoxMenuItem1ActionPerformed
```
eholum: Instead of having a separate method in the plotterview, could the FunctionModel be part of the VisualizerDataModel? Then this could set a preference in the stilplotter or something that there is a FunctionModel to plot, and it could get it when creating the MapEnvironment in the StilPlotter?

jbudynk: Good point. Although now you got me thinking this should be a part of the SedModel too, since for beta1 at least, we're assuming models are associated with a unique SED.

The VisualizerDataModel is setup to define multiple selected SEDs in the future. What if we say a SED can have multiple models? This would be a use-case of Tier 3 requirement 2.2.c.1 in the BRD (plotting individual model components along with the composite model). In that case, the VisualizerDataModel would have to handle a LinkedList of LinkedLists of FunctionModels, which definitely doesn't seem like the right way to handle this case... Maybe this could be a "CompositeFunctionModel" which is just a list of FunctionModels.

Is this out of the scope of this PR, or should I try to tackle it here?

eholum: I like that idea! Then we could just call from the StilPlotter to get the function models from the SedModel, assuming that the fitting tool puts them there, too, or with the SED - as they're both present.

My initial thinking was that a FunctionModel class would natively support Lists of SegmentStarTables each representing a fitted SED (though to start this can just be one), then just use a LayerModel as the plot layer, rather than adding another type of Layer (but that's in https://github.com/ChandraCXC/iris/pull/303)

Long story short I don't think we'll need to add another class to handle multiple fits for a single SED, it can be done with what you've got going so far.

jbudynk: Thanks for the explanation. So if I'm to rebase this PR on top of #303, and I use the LayerModel to plot the FunctionModel, I'll have to add support for line plots to the `LayerModel`, right? The FunctionModel uses a line plot instead of a scatter plot, for example, and the LayerModel only plots points and error bars.

eholum: Yeah pretty much. At somepoint we had an enum for LayerType and I vaguely remember making one of the line... but that may have been months ago.

olaurino:
> What should happen if a user overplots a relatively large model on a short SED? I.e., if the model's x range and/or y range >> SED's x and y range

I am not sure I understand the question. The model is evaluated at the Xs defined in the Sed (or, more specifically, in the `SedModel`), so the Sed and the Model are always of the same length.

jbudynk:
Ah, that's right. Iris 2.1 only plots a template or model within the data
range... alright, so I guess we shouldn't worry about this.

I imagine it would be useful down the road to allow one to view the full
model over a user-defined range. I'm thinking evaluated models or fitted
templates in a relatively short x-range. For example, fitting a template
defined from 0.1 um to 1e8 um in the IR (1 - 1e4 um), but seeing how the
template looks in the optical and radio wavelengths.

But that's down the road :)

On Thu, Jun 23, 2016 at 4:21 PM, Omar Laurino <notifications@github.com>
wrote:

> What should happen if a user overplots a relatively large model on a short
> SED? I.e., if the model's x range and/or y range >> SED's x and y range
>
> I am not sure I understand the question. The model is evaluated at the Xs
> defined in the Sed (or, more specifically, in the SedModel), so the Sed
> and the Model are always of the same length.
>
> —
> <https://github.com/ChandraCXC/iris/pull/304#issuecomment-228171731>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AF8elpw9Sf_icPjoYHubGXVZwyP3Bw57ks5qOurjgaJpZM4I82i8>
> .
>



-- 
Jamie A. Budynkiewicz
Cambridge, MA 02140

jbudynk:
rebasing on #303. original tip was 43d09ee. 

related commits are from 607f30e to 631c937

jbudynk:
Rebased on master (original tip was de09446)

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
```diff
@@ -0,0 +1,151 @@
...
+    private String functionColor = "red";
+    private Double functionDash = Double.NaN;
+    private Integer functionThickness = 1;
+
+    public FunctionModel(SedModel sedModel) {
+        
+        this.sedModel = sedModel;
+        // concat IrisStarTable into StackedStarTable
+        StackedStarTable stackedTable = new StackedStarTable(sedModel.getDataTables(), new SegmentColumnInfoMatcher());
+        stackedTable.setName(sedModel.getSedLayerModel().getSuffix());
```
eholum: Maybe add something to this, like + " Fit" or something. O/w they'll probably be some confusion in the legend (if these show up in the legend)

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ *
+ * Evaluated model preferences. This describes the STIL preferences for plotting 
+ * an evaluated model.
+ */
+public class FunctionModel {
+    
+    public static final String DEFAULT_FUNCTION_COLOR = "red";
+    public static final String RATIO = "ratio";
```
eholum: Can you make these "Ratios" and "Residuals" to match whats in the plotterview?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
```diff
@@ -0,0 +1,151 @@
...
+                    + "Must be \"ratio\" or \"residual\"");
+        }
+        
+        layer.setShowErrorBars(false);
+        layer.setShowMarks(true);
+        
+        return layer;
+    }
+    
+    public void setSedModel(SedModel sedModel) {
```
eholum: Should this function change the sortedStarTable?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
jbudynk: A suffix gets added in the `try` statement below, at line 66. The suffix is "_MODEL"

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
```diff
@@ -0,0 +1,151 @@
...
+        this.sedModel = sedModel;
+    }
+    
+    public SedModel getSedModel() {
+        return this.sedModel;
+    }
+    
+    public void setSortedStarTable(SortedStarTable table) {
```
eholum: This is fine for now, but there's still some thinking to do here - like when users mask/hide points it will change the length of the base startables and mess up the row sorted values. I think this is just going to have to be recomputed every time the SedModel changes.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/LayerModel.java
```diff
@@ -117,6 +134,14 @@ public int getNumberOfLayers() {
             addErrorFields(suffix + ERROR_SUFFIX, preferences);
         }
         
+        if (showLine) {
+            addLineFields(suffix, preferences);
+        }
+        
+        // if neither marks nor errors are shown, just add the common fields.
+        if (preferences.isEmpty())
```
eholum: Use braces!

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerDataModel.java
```diff
@@ -189,6 +199,9 @@ public synchronized void setSelectedSeds(List<ExtSed> selectedSeds) {
             else {
                 newSedModels.addAll(sedModel.getLayerModels());
             }
+            // set a FunctionModel
+            // TODO: handle setting multiple function models
+            //newFunctionModel.add(sedModel.getEvalModel());
```
eholum: When this is uncommented this should probably be in an 'if' block if it has a FunctionModel.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
eholum: Ah, perfect.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
jbudynk: Yeah, that makes sense. Adding it now. It'll just create a new SortedStarTable from the updated SedModel.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
jbudynk: I see what you're saying. I can create a new SortedStarTable from the SedModel.

jbudynk: Would it make more sense to have a "updateSortedStarTable()" method instead?

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedModel.java
```diff
@@ -226,6 +228,21 @@ boolean removeSegment(Segment seg) {
     }
     
     /**
+     * Attaches an evaluated model to the Sed.
+     * @param model
+     */
+    public void setFunctionModel(FunctionModel model) {
```
eholum: I probably wouldn't even bother with a setter function here, just the getter - and I might have the getFunctionModel method recompute the actual sortedStarTable each time it's called. That way it gets the most up-to-date IrisStarTables possible.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
eholum: I would actually probably hide the entire notion of the underlying sortedStarTable all together. The FunctionModel is entirely tied to the SedModel/List of IrisStarTables in the SedModel, so I wouldn't give callers access to this object's underlying functionality by allowing them to set the sorted star table. It's the Function Model's job to turn the SedModel into a useable Layer, so I would just black box all that.

For now I would just have the setSedModel function.

jbudynk:
Rebased on master. Original tip was 888996c.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.java
```diff
@@ -248,6 +256,7 @@ private void initComponents() {
         mntmAutoFixed = new javax.swing.JCheckBoxMenuItem();
         mntmGridOnOff = new javax.swing.JCheckBoxMenuItem();
         mntmCoplot = new javax.swing.JMenuItem();
+        mntmShowModel = new javax.swing.JCheckBoxMenuItem();
```
eholum: Random, but can this go above the coplot option? Just so all the checkbox preferences are first.

eholum: Or even make this as a checkbox down by the residuals or something? Or if we're always plotting the fit if it's available I guess we don't even need it yet...

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
jbudynk: sure thing

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedModel.java
jbudynk: I agree that this makes sense. Some issues I'll have now is hiding/showing the model on the plot, setting up a dummy FunctionModel to plot on the PlotterViewer (which is fine since we're getting rid of it anyhow), and creating unit tests. I'll need a little time to update these things.

eholum: Yeah no worries, there's definitely still some tweaking left to do in that menu bar, what with items that have to be removed an all. Maybe just leave it as is for now and we can tweak the UI after the functionality is there.

eholum:
I think this is in a good place. When the builds pass go ahead an merge it and we can move forward with some of the other tasks.

jbudynk:
I'm merging this into master once the tests pass. I want to address the other issues @eholum mentioned here, especially on removing the `setFunctionModel()` method from `SedModel`, as this impacts other parts of the code, in parallel to the Plotter/Fitting tool integration tasks.

PR #298, opened 2016-05-26
Fitting Tool Evaluate Model

olaurino:
Resolve chandracxc/iris#73.

# Description
This PR introduces model evaluation features to the `FitController` class.

# Notes
I made some clean up in the relative `sherpa-samp` call, which then required a good simplification in the `SherpaIntegrator` class.

I am building on top of #292 because of some infrastructure requirements. The only commit relevant to this PR is e268e39.

PR #305, opened 2016-06-23
Model evaluation startable

olaurino:
This PR modifies the model evaluation method created in #298 to return a `SegmentStarTable`, which can be integrated in the visualizer to plot model functions. It also accepts a `SedModel` instance as an argument, as this is the model class used by the visualizer internally.

Units conversion should be properly handled by the method and its dependencies. When reviewing, please pay special attention to the tests: they should make sure the conversion is properly handled, but we need careful validation that I did the right calculations (I used astropy to validate the conversions, but I might have made the same mistake twice ;).

Also, you are welcome to suggest error handling test cases to add.

Note that this PR is based off of #303 for convenience, but only 3db91cf is relevant for this PR.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/fitting/FitController.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/fitting/FitController.java
@@ -8,6 +8,7 @@ import cfa.vo.iris.sed.stil.SegmentStarTable;
 import cfa.vo.iris.units.UnitsManager;
 import cfa.vo.iris.visualizer.preferences.SedModel;
 import cfa.vo.iris.visualizer.stil.tables.IrisStarTable;
+import cfa.vo.sedlib.Sed;
 import cfa.vo.sherpa.ConfidenceResults;
 import cfa.vo.sherpa.Data;
 import cfa.vo.sherpa.FitResults;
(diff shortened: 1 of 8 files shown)


olaurino:
Rebasing, original tip: 3db91cf

olaurino:
Note that I think we should have delegating methods from `IrisStarTable` to `SegmentStarTable` to get the *raw* data, but that's hardly a priority.

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FitController.java
```diff
@@ -201,12 +205,61 @@ public FitConfiguration loadJson(InputStream is) throws IOException {
...
+        String xUnit = sedModel.getXUnits();
+        String yUnit = sedModel.getYUnits();
+
+        UnitsManager uManager = Default.getInstance().getUnitsManager();
+
+        for (IrisStarTable table: sedModel.getDataTables()) {
+            double[] x = table.getSpectralDataValues();
+            double[] xStandardUnit = uManager.convertX(x, xUnit, SherpaClient.X_UNIT);
+            double[] yStandardUnit = client.evaluate(xStandardUnit, sedModel.getFitConfiguration());
+            double[] y = Default.getInstance().getUnitsManager().convertY(yStandardUnit, xStandardUnit,
```
eholum: This looks right to my eyes/writing it down. Ditto the tests. Not sure if that's worth that much but hey.

eholum:
Looks correct to me. Merge it and we can use it!

PR #306, opened 2016-06-23
Add Residuals Plot to StilPlotter

eholum:
Basic implementation of the secondary plotter - in appearance only! Need to have the FunctionModel infrastructure in place and then should be able to add layers to the Residuals plot display in the setupResidualMapEnvironment method.

I spent some time trying to figure out how to fix the planeaspects together so that they would zoom at the same time, but it's non-trivial so leaving it for a later time.

For now the residuals will just plot the same first layer that's available in the regular plotter - that will change going forward.

This is also based off of https://github.com/ChandraCXC/iris/pull/303, relevant first commit is  	c0d6005, then the last 4.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
@@ -88,7 +88,7 @@ public class FunctionModel {
             layer.setY(SegmentColumn.Column.Residuals.name());
         } else {
             throw new IllegalArgumentException("Unrecognized residual type. "
-                    + "Must be \"ratio\" or \"residual\"");
+                    + "Must be \"Ratios\" or \"Residuals\"");
         }
 
         layer.setShowErrorBars(false);
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
@@ -16,13 +16,12 @@
 package cfa.vo.iris.visualizer.preferences;
 
 import cfa.vo.iris.sed.stil.SegmentColumn;
+import cfa.vo.iris.sed.stil.SegmentColumn.Column;
 import cfa.vo.iris.visualizer.plotter.LayerType;
-import cfa.vo.iris.visualizer.stil.tables.SegmentColumnInfoMatcher;
 import cfa.vo.iris.visualizer.stil.tables.SortedStarTable;
-import cfa.vo.iris.visualizer.stil.tables.StackedStarTable;
+
 import java.io.IOException;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import uk.ac.starlink.table.StarTable;
 import uk.ac.starlink.ttools.jel.ColumnIdentifier;
 
 /**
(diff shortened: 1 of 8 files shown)


eholum:
Ah, well apparently the GridBagConstraints don't seem to work in Java 8, or at least not on my mac. Hooray.

eholum:
Rebased and added added new commit for adjusting layout.

eholum:
Rebased. Previously at 1e59fd9.

eholum:
+ Added residuals to the StilPlotter by uncommenting the FunctionModel stuff from the DataModel. 
+ Added additional verification on the residuals layer in the function model unit test. 
+ Added additional checks to make sure the correct columns are present before plotting model functions.
+ Moved the SedModel out of the FunctionModel - I guess it really only needs a star table and can go from there.

eholum:
This now resolves

https://github.com/ChandraCXC/iris-dev/issues/106
https://github.com/ChandraCXC/iris-dev/issues/109

There's definitely still some weirdness in the relationship between the fitting tool and plotter, but once the values are there they should show up.

olaurino:
Is the weirdness you are mentioning related to the fact that the fitting tool is not using `SedModel` (if not in one method)? I think that could be fixed in the scope of other integration tasks, like getting and setting the fitting ranges.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/VisualizerDataModel.java
```diff
@@ -291,15 +295,15 @@ public synchronized void setSelectedStarTables(List<IrisStarTable> newStarTables
...
-//    public synchronized void setFunctionModels(List<FunctionModel> newFunctionModels) {
-//        List<FunctionModel> oldFunctionModels = functionModels;
-//        this.functionModels = ObservableCollections.observableList(newFunctionModels);
-//        pcs.firePropertyChange(PROP_FUNCTION_MODELS, oldFunctionModels, functionModels);
-//    }
+    public List<FunctionModel> getFunctionModels() {
+        return functionModels;
+    }
+    
+    public synchronized void setFunctionModels(List<FunctionModel> newFunctionModels) {
```
jbudynk: What is the reason for setting the FunctionModels from the VisualizerDataModel? This would make the list of FunctionModels independent from the SedModels' functions, correct?

eholum: That's an excellent point. I want the setter function for the property, and someday it might not be tied to the SedModels, but since this is tied to setting the List of selectedSeds I'll make this private for now. In the future this can be managed separately.

eholum:
It definitely could be, I haven't spent too much time playing with it. Right now there's not really any good logic in determining if a Fit is attached to an SED other than just looking and seeing if the model_value column is available. That logic needs to be tighter and could be expanded w.r.t the fit notification task we have out.

iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/FunctionModel.java
```diff
@@ -94,49 +90,64 @@ public LayerModel getResidualsLayerModel(String residualType) {
             throw new IllegalArgumentException("Unrecognized residual type. "
                     + "Must be \"ratio\" or \"residual\"");
```
jbudynk: this is a super minor typo, but I forgot to change this one in the #304 PR. The error message should be "Must be \"Ratios\" or \"Residuals\"" to go along with the naming scheme here.

eholum: Ah, good catch. Will fix that.

jbudynk:
The changes look good. I give it a thumbs up to merge. Nice job cleaning up the leftovers of PR #304, too.

eholum:
Sounds good, will merge it after the build succeeds

eholum:
Merged #306

PR #307, opened 2016-06-24
Enable Panning Buttons for the Plot Window

eholum:
The little arrows in the plot window didn't do anything - so I added a simple notion of panning to the StilPlotter.

Basically for linear scales the axis moves by 1/4 the size of the axis. So panning on an axis of length 10 will move the axis by 2.5 in the specified direction.

Log scales will pan by a factor of 1/3, in either the positive or negative direction. I tried to figure out what Iris2.1 does, but it's panning is rather odd, so hopefully that makes sense.

Easy to change later on, I just wanted this functionality to be there.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.form
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/PlotterView.form
@@ -375,6 +375,9 @@
                   <Dimension value="null"/>
                 </Property>
               </Properties>
+              <Events>
+                <EventHandler event="actionPerformed" listener="java.awt.event.ActionListener" parameters="java.awt.event.ActionEvent" handler="upActionPerformed"/>
+              </Events>
             </Component>
             <Component class="cfa.vo.iris.visualizer.plotter.JButtonArrow" name="down">
               <Properties>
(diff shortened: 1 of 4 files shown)


olaurino:
Tested it and tried to stress it (not that I expected it to fail). Merge it.

I was hoping we could finally get nicer markers for the points. Can we just render points as, ahem... *points*? The current circles are interfering with the error bars and it doesn't look right.

Not that it's important in this PR, but since we're there :)

eholum:
Oh yeah, that's easy to do.

Just got to change it to one of these:

http://andromeda.star.bris.ac.uk/starjavadocs/uk/ac/starlink/ttools/plot/MarkShape.html

I don't have a preference, so whatever you would prefer we can do.

On Fri, Jun 24, 2016 at 4:42 PM, Omar Laurino <notifications@github.com>
wrote:

> Tested it and tried to stress it (not that I expected it to fail). Merge
> it.
>
> I was hoping we could finally get nicer markers for the points. Can we
> just render points as, ahem... *points*? The current circles are
> interfering with the error bars and it doesn't look right.
>
> Not that it's important in this PR, but since we're there :)
>
> —
> <https://github.com/ChandraCXC/iris/pull/307#issuecomment-228456896>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AHmNVVTCgws4YGI96bXRGDkC6WzhIp2Nks5qPEEwgaJpZM4I-GTW>
> .
>



-- 
Erik Holum

eholum:
Oh and, errors can be:
    none
    lines
    capped_lines
    caps
    arrows
    ellipse
    crosshair_ellipse
    rectangle
    crosshair_rectangle
    filled_ellipse
    filled_rectangle

Just need to change it in the default preferences object.

PR #308, opened 2016-07-15
iris-dev#77 Plot Evaluated Models

eholum:
Resolves 
https://github.com/ChandraCXC/iris-dev/issues/76
https://github.com/ChandraCXC/iris-dev/issues/77

Adds basic support for plotting the evaluated models after a fit. After a fit is completed the fitting tool immediately calls to evaluate the sed asynchronously from the visualizer preferences. I added a singleton holder, IrisVisualizer, for objects that need to be shared between the fitting tool and the plotter/MB. At this point it's pretty simple, but will need to be expanded for future integration of the two components.

A few points:
  + The models are evaluated asynchronously, and the datamodel will be updated after the evaluation completes.
  + As of now there is no notion of validation on the models. That will come in a later commit.
  + I fixed the unit tests in the visualizer, they were dependent on the serialization of the the SED in the swing EDT, now they are not.
  + Updated the color palette so that external callers can specify a starting color. Model functions will start at red then move around from there.
  + Changed the appearance of residuals/ratios to match whatever shape is currently in the plotter. Right now that is open circles, but easily changed.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/fitting/FitController.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/fitting/FitController.java
@@ -3,14 +3,11 @@ package cfa.vo.iris.fitting;
 import cfa.vo.iris.fitting.custom.CustomModelsManager;
 import cfa.vo.iris.fitting.custom.DefaultCustomModel;
 import cfa.vo.iris.fitting.custom.ModelsListener;
-import cfa.vo.iris.sed.ExtSed;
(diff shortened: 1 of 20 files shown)


eholum:
Oops, forgot to commit a file. Added and rebased. Previous head at 008028010c43e734bf74e9af6041753196d5fdf0.

jbudynk:
ok to merge

eholum:
merged

PR #309, opened 2016-07-15
iris-dev#111 Model Validation

eholum:
Resolves
https://github.com/ChandraCXC/iris-dev/issues/111

Validation uses hashCodes built on the spectral, flux, and error values in a SegmentStarTable. The logic being that a fit will be valid as long as those values haven't changed. The SedModel uses those hashCodes to produce a "hashCode" for all the segments in the SED. Note that I call it a version rather than a HashCode, as I don't really want to overwrite the SedModel's hashCode function as it's pretty heavily utilized already in the Visualizer.

Right now the default behaviour is to show a warning message once when a function model is no longer valid for the given SedModel. Once the user has acknowledged the warning it will not be shown again until the SED version number changes.

The Version numbers are recorded in two places: The FitConfiguration whenever .fit is called in the FitController, and in the SedModel whenever .evaluateModel is called in the FitController.

The FitConfiguration now has a version number attached to it, though this change isn't currently using it. I can easily add an additional check in the StilPlotter if we think it should go there. Otherwise I would like to see something in the Fitting Tool also doing a check to see if the FitConfiguration is still valid for the SED. There's definitely more behaviour we could add in the Fitting Tool but I won't mess with it until the GUI changes are in place.

### DIFF
--- a/iris-visualizer/src/test/java/cfa/vo/iris/visualizer/preferences/SedPreferencesTest.java
+++ b/iris-visualizer/src/test/java/cfa/vo/iris/visualizer/preferences/SedPreferencesTest.java
@@ -18,6 +18,9 @@ package cfa.vo.iris.visualizer.preferences;
 
 import static org.junit.Assert.*;
 
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
 import static cfa.vo.iris.test.unit.TestUtils.*;
 
 import org.junit.Test;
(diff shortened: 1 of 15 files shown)


eholum:
Added a change to remove unused menu items from the plotter - since we're not using them and aren't going to be using them, better to remove them from the UI.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/preferences/SedPreferencesTest.java
```diff
@@ -142,4 +145,52 @@ public void testAddMultipleSegments() throws Exception {
...
+        sed.addSegment(segment2);
+        model.refresh();
+        int h3 = model.getVersion();
+        assertFalse(h2 == h3);
+        
+        // Remove 2nd segment
+        sed.remove(segment2);
+        model.refresh();
+        int h4 = model.getVersion();
+        assertTrue(h2 == h4);
```
jbudynk: so does this mean if you (1) fit a SED, (2) add a segment, (3) ignore the warning (ie don't refit or re-evaluate), then (4) remove the lastly added segment, that I shouldn't see the warning pop up?

jbudynk:
The code looks good. 

One comment I have is when I add a new segment to the sed, could the warning message pop-up after the the "Success" message pops up, and the Load Sed Frame closes?  I haven't look at how this could be done yet to be honest.

Also, I think it'd look nicer if the warning message popup was thinner; if there is a line break in the message, it'd make it look better. But that's also just a personal preference on my part.

iris-visualizer/src/test/java/cfa/vo/iris/visualizer/preferences/SedPreferencesTest.java
eholum: Unfortunately not - though that is adjustable.

If you ignore the warning then the model version is moved up to the current state of the SED. This prevents the warning from popping up every time to refresh the plotter, but will make it pop up any time you change the SED, even if it goes back to being valid.

eholum:
Oooph, right now I don't think theres a system for sequencing popup windows, so I'm not sure how simple that would be. We could always move the success message up a little bit earlier in the code, but that's no guarantee.

Easy enough to add a line break. I'll do that.

eholum:
Any thoughts on merging this guy?

jbudynk:
I did a quick test and noticed an issue:

If I change the units on the plot, I get that warning message that the SED has changed. 

Looking at the PR description, it makes sense that this warning would pop up. Changing the units on the plot changes the segment values in the VisualizerDataModel, right? However, switching the units shouldn't be considered a segment value change.

eholum:
That's a really good point. I need to think about how to update everything when the plotter units change. The hashCode is computed from the values in the SegmentStarTable, and the values change in the table when changing units, but I can probably put in some special logic in the SedModel for when units are changing. Update in a  min.

jbudynk:
The latest change looks like it fixes the issue with changing units. The break in the warning message looks good too. Thanks for getting the changes in.

eholum:
Added change for updating the model and fit versions on units changes. It's less than ideal, but as the units are set directly in the datamodel from potentially multiple locations, I just added a separate method in the datamodel for bringing all version numbers up to the latest. That should address it.

eholum:
Merged #309

PR #311, opened 2016-07-18
Iris dev 78 get fitting ranges

jbudynk:
Addresses:
ChandraCXC/iris-dev#78 - get fitting ranges from plotter
ChandraCXC/iris-dev#79 - set fitting ranges

# Description

Fitting ranges are set by clicking start and end points on the plotter. On the Fitting Tool, the user clicks "Add Range", then clicks on a start and end point to denote the fitting range. Multiple ranges can be defined. Fitting ranges can be cleared by clicking "Clear Ranges."

Ranges are stored in a List<FittingRange> in the FitConfiguration. One can access the ranges with FitConfiguration.getFittingRanges(). FittingRange is a simple object that has getters/setters for the start and end points, as well as the spectral unit.

All ranges are converted to Angstroms in FitConfiguration.addFittingRange(). The order of the start and end points is sorted in addFittingRange() as well.

If the Iris Visualizer is not open or visible, and "Add Range" is clicked, then a warning message pops up asking the user to open the Visualizer in order to set the fitting ranges.

Caveats:

 - Currently, there is no way from the GUI to remove a selected range; one has to remove all the ranges to remove one of them.

TODOs:

 - there is no test to check that the MouseXRangesClickedListener works as expected. While I've tested that this works by hand often, I'd feel better writing an automatic test that opens both the Fitting Tool and Visualizer, adds a range by clicking on two points on the plotter, then checks that the FitConfiguration contains the expected fitting range.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingRangesFrame.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingRangesFrame.java
@@ -343,6 +343,9 @@ public class FittingRangesFrame extends javax.swing.JInternalFrame {
             Logger.getLogger(FittingRangesFrame.class.getName()).log(Level.SEVERE, null, ex);
         }
 
+        // refresh the plot to show new fitting range
+        preferences.getDataModel().refresh();
+        
         // update table
         updateTable();
     }//GEN-LAST:event_addRangeButtonActionPerformed


jbudynk:
Note that the Travis build for Java 1.8 is failing at `testSetFittingRangesNoPlotter(cfa.vo.iris.visualizer.FittingToolComponentTest)`. It passes on Java 1.7, and on my workstation (which is also Java 1.7). I will need to investigate why this is happening.

I will also need to rebase my changes on top of PR #310 since it updates the Fitting Tool layout.

iris-common/src/main/java/cfa/vo/iris/fitting/FittingRange.java
```diff
@@ -0,0 +1,65 @@
...
+package cfa.vo.iris.fitting;
+
+/**
+ *
+ * Simple object for holding the start and end points of a fitting range.
+ */
+public class FittingRange {
+    private double startPoint;
+    private double endPoint;
+    private String xunit;
```
eholum: can you make this an XUnit object?

iris-visualizer/src/main/java/cfa/vo/iris/fitting/FittingMainView.java
```diff
@@ -447,9 +474,30 @@ private void doFit(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_doFit
...
 
+    private void addFitRangeButtonActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_addFitRangeButtonActionPerformed
+        // if there are no SEDs, don't set the fitting ranges
+        if (this.getSedModel() == null) {
+            return;
+        }
+        
+        // if the PlotterView isn't up, ask the user to open it
+        MouseXRangesClickedListener listener = (MouseXRangesClickedListener) this.preferences.getMouseListenerManager().getListener(MouseXRangesClickedListener.class);
+        if (listener.getPlotterView() == null || !listener.getPlotterView().isVisible()) {
```
eholum: null check the listener, too, since that method specifically returns null

iris-common/src/main/java/cfa/vo/iris/fitting/FittingRange.java
jbudynk: cfa.vo.iris.sed.quantities.XUnit

or

cfa.vo.iris.units.XUnit?

jbudynk: (and I agree it should be changed to one of those. I'm veering towards cfa.vo.iris.sed.quantities.XUnit as that's got all the enums we use. I've got working versions for both objects)

jbudynk: I made `FittingRanges.xunit` a cfa.vo.iris.sed.quantities.XUnit.

eholum:
Two questions:
1) How are the ranges clicked displayed to the user? Or the set ranges?
2) When the 'fit' button is clicked, can the plotterview come to the front if it's open? Also, is there a way to have some kind of window or something shown that shows what values have been chosen as min and max?

jbudynk:
Rebasing. Original tip was 8c126fa.

jbudynk:
I've added a frame for viewing / editing the fitting ranges. One can now add a specific fit range to the fit configuration, and remove a selection of the fit ranges instead of removing all of them. This "fitting range manager" can be accessed from the menu bar under "Edit." 

GUI tests have also been added. Note that I still don't have a test for mocking fit ranges added from the Plotter, but I've checked them by hand.

Caveats:
  - If adding ranges from the Plotter in Co-plot mode, the fit ranges will be added to each co-plotted SED.
  - If a user resets the plot in the middle of adding ranges from the plotter, the user will have to click somewhere on the plot to finish defining the ranges.

eholum:
Nice, I really like the look of that window. What about removing the add button on the fitting tool, and just putting in an add/edit fitting ranges button that opens that up? Maybe?

jbudynk:
Yeah, I think that'd be nice, too. The reason I didn't do that in the first place was so there didn't have to be so many windows popping up, but I think the manager is condensed enough for easy use. @olaurino, do you have any preference?

olaurino:
Not really :)

jbudynk:
I can't get the Java 1.8 build to pass on Travis. Also, Travis has been having issues all day, making my builds take over 30 minutes to run. This branch passes on my laptop which uses Java 1.8 (1.8.0_40). I will rebase this off of master, but I am unsure on what to do after.

jbudynk:
I merged PR #316 into this branch, which has the logic for showing the fitting ranges on the Plotter.

jbudynk:
Rebasing on master. Original tip was 12027ac.

jbudynk:
Cleaned up messy rebase. (bad) original tip was b033bee.

olaurino:
I removed commits ce13310 and 12d2ab8, which are now 86bf859 and 3f86044.

PR #313, opened 2016-07-21
Add Vizier/CDS plugin

olaurino:
Fix chandracxc/iris-dev#84.

This PR introduces a rather simple plugin for fetching data from the Vizier photometry/SED service.

It's based on an early prototype which was revamped to include error handling, few tests, and to improve the GUI, which is still very simple.

As for testing, I spent quite some time trying to figure out why I could not have different functional tests for the whole GUI, then I gave up. Most likely, the way a Swing `Application` works some static data is leaking when invoking the main method of the an extending application, so that if the main method is invoked more than once (as it happens during tests), the GUI does not work properly. At runtime, the exit logic actually terminates with a system exit, but that has to be short-circuited during tests.

The solution for now is to have only one functional test with all the tested interactions.

I based this PR off #310 to get some test changes, but the relevant commits start with 	d1fde2a.


olaurino:
Rebasing. Original tip was at 0ca09b0.

olaurino:
Rebasing to remove spurious commit from a different PR (argh!). Previous tip was eae8b8d.

jbudynk:
Looks good on Linux. Though I would suggest adding more padding around the box edges.

On OS X though, I can't change the size of the window, I can't see what I've typed in for the search radius (because the box is very narrow), and I can't see a "Submit" button.

Viewing this on the Plotter reminds me we need to implement the legend RFE soon too :)

jbudynk:
The layout looks much better on OS X. Looks fine on Linux too.

olaurino:
Are we good to merge this?

PR #314, opened 2016-07-22
Fit masked data

eholum:
Replaces #312 - opening against master.

Resolves
ChandraCXC/iris-dev#112

Added support for the fitting tool to get masked data instead of all the data when computing a fit. Additional unit/functional tests at several layers verify the changes.

One note, I made an interface for the star table that we use for plotting so that we can interchangeably use IrisStarTables and SegmentStarTables, when required.

Also, note that this is branched off model-validation for ease of viewing.

In order to satisfy the conditions of the FittingFunctionalIT, we always use the original Spectral, Flux, and Flux Error values when fitting, rather than the data currently in the plotter.

EDIT:
Also included the changes for optimizing construction of SegmentStarTables.


olaurino:
I quickly reviewed the code, and I am OK with the changes.

PR #315, opened 2016-07-25
Show and Hide Legend

eholum:
Adding jcheckbox for showing and hiding the legend on the plotdisplay. Show and hiding does not alter the plot's aspect.

Also disabling the filter expression panel in the metadata browser, for now.


jbudynk:
Looks good. I tested it out with some ASDC data, and it works as expected.

PR #316, opened 2016-07-25
Plot Fitting Ranges

eholum:
Opening against #311 since this includes all of those changes.

Plots fitting ranges as a new layer in the stil plotter. There's a significant TODO in the StilPlotter's construction of the plot display, in that we don't have access to the aspect's bounds prior to constructing a plot, so we just do it twice, if necessary. This keeps the fitting ranges from affecting the default view ranges on the plotter.

Ranges are always plotted at 10% from the y-axis minimum.


jbudynk:
The ranges look fine, but it doesn't look like it's plotted at 10% the minimum y-value. On a linear scale it looks close, but on a log scale, it looks more like 10% the maximum. However for now I think it's fine.

I also noticed that if I pan the plotter so that the midpoint of the fitting range falls off the edge of the plotter, the blue line disappears. This one is a little more annoying, but the issue could lie in STILTS and not Iris. Maybe using a different mark type would change the behavior; but I'd add it as an item to ask Mark Taylor for now.

jbudynk:
I'm OK with merging this into iris-dev-78.

PR #319, opened 2016-08-02
Fix Assorted Bugs

eholum:
Opening this now to prevent any one PR from containing too many commits in one go. This PR address the following issues:

+ Co-plotting an empty SED throws exception - the dataModel will skip adding empty SEDs
+ Show legend not correctly bound on plot window - updated the plotterview method to include the show legend option.
+ Masking points throws an arrayindexoutofbounds exception when re-plotting with model functions - since the model functions are based on all points, but were trying to use masked tables - fixed that in the SedModel
+ Fitting ranges don't show up at the right spot on log-scales - address that in the StilPlotter
+ Removed Unused menu items from the Metadata Browser.

EDIT:
+ Fixed issue with grid lines and a fixed plot.
+ Fixed error related to sorting/masking in the metadata browser.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
+++ b/iris-common/src/main/java/cfa/vo/iris/sed/stil/SegmentStarTable.java
@@ -25,7 +25,6 @@ import java.util.TreeSet;
 import java.util.UUID;
 
 import cfa.vo.iris.sed.ExtSed;
-import cfa.vo.sherpa.SherpaClient;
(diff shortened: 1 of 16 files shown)


eholum:
Added change for making plot marks filled diamonds, error bars larger. Easy to tweak going forward if this still isn't good.

eholum:
Rebased over release/3.0b1

olaurino:
Merged, although in bits and pieces.

PR #320, opened 2016-08-03
round of bug fixes

### LINKED ISSUE #259: Saving a Segment: All segments besides the selected one are removed from the SED Builder view
To reproduce:

Load a 5 NED SEDs (for example: m81, m82, m83, m84, m85).
Select the 3rd one and click the "Save" button.

The other segments will disappear from the Segments table.

However, if you switch to a another SED and back to this one, you'll see all the original segments are in the table.

olaurino:
This PR fixes the following bugs.

  * Fitting tool GUI should now be consistent on all platforms.
  * resolve #259.
  * make interpolation work again (it was broken by a refactoring).
  * integration tool copies model expression rather than overwriting it.
  * integrator now properly handles empty SED.
  * fitting tool has proper icon and tooltip.
  * column names in confidence table have been changed, and the parameter value added for convenience.
  * vizier client explicitly handles connection errors.

I also added an ignore to a test that (almost?) always fails on Travis.

I plan to add some tests later.


eholum:
Confirmed fixes on my linux machine. I have a fix on my branch for that test, not sure what the issue in travis is but it seemed to have worked - still and issue with the FT component test on that branches builds.

Merge this and I can rebase my stuff.

Issue #99, opened 2015-03-18
SED STACKER - Add "Remove" option when user right-clicks Stack in "Open Stacks"


jbudynk:
I realize that there is the “Management” section that is the place where all commands about the stacks are located, but in the case a reorganization of the layout of the sections in the main SED Stacker frame is off the table (#97), I would add at least the “Remove” action in the contextual menu accessed by right­-clicking on the name of the open stacks.

Issue #98, opened 2015-03-18
SED STACKER - GUI: typos. "Managment" section and remove "!" in "Stack!"


jbudynk:
there is a typo in the title of the “Managment” section and I would remove the exclamation point from the “Stack!” button for consistency with the other buttons.

PR #323, opened 2016-08-10
Fix miscellaneous sedstacker bugs / rfe's

jbudynk:
This PR addresses the following sedstacker-related bugs and RFE's:

- Segment name on Plotter legend is a messy UID of that segment layer. Now, the segment Target name of the stacked segment is set to the name of the Stack.
- Typo fixes on sedstacker GUI (#98).
- Add a "Remove" option when the user right-clicks on a Stack in the "Open Stacks" list (#99).


jbudynk:
Rebased on release/3.0b1 branch. Original tip was 97a1a35.

PR #325, opened 2016-08-10
Fitting range updates

jbudynk:
This PR addresses the following issues:

- Fitting ranges were not applied to the data before being sent to Sherpa for fitting.
- Clicking the "Add ranges" button from the Fitting Tool caused multiple fitting ranges windows to pop up.

Now, fitting ranges are applied to the data before being sent to Sherpa (done in the FitController), and only one fitting ranges window is open at a time.

Note: there is still an issue with the units the data is fit in. I would like to fix this issue in this PR if no one else addresses it.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedModel.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedModel.java
@@ -182,7 +182,7 @@ public class SedModel {
     // TODO: Make this better! e.g., should have some mechanism to just have a stacked
     // IrisStarTable that would allow us to more easily and efficiently extract data. Still, 
     // performance wise much better than sedlib.
-    public void getFittingData(final Data data, boolean includeMasked) {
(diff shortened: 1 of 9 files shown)


iris-visualizer/src/main/java/cfa/vo/iris/visualizer/preferences/SedModel.java
```diff
@@ -182,7 +182,7 @@ public FunctionModel getFunctionModel() {
     // TODO: Make this better! e.g., should have some mechanism to just have a stacked
     // IrisStarTable that would allow us to more easily and efficiently extract data. Still, 
     // performance wise much better than sedlib.
-    public void getFittingData(final Data data, boolean includeMasked) {
+    public void setFittingData(final Data data, boolean includeMasked) {
```
olaurino: I don't think this change was intended, was it?

jbudynk: it was -- i thought "set" made more sense than "get"... but yeah, it was a silly minor change. I can change it back

PR #326, opened 2016-08-10
Address iris-dev#56: Filter Expressions in MB (double-valued columns only)

jbudynk:
This PR addresses part of ChandraCXC/iris-dev#56. It implements a filter expression validator and builder, and then selects the rows in an IrisStarJTable which comply with the filter expression.

There are tests for parsing the filter expressions. Currently, there are no tests for integration with the metadata browser. I did run a couple tests by hand, however, to make sure basic filters work as expected. More tests are needed before merging this PR.

Also, if we add this to the 3.0b2 release, we will need to address the licensing issue with javaluator.


jbudynk:
caveat (or bug?): can't use negative scientific notation. For example, `$2 > 1e-5` throws an error. However, `$2 > 1e5` works fine. I'll see if javaluator has notes on writing negative exponents for scientific notation...

jbudynk:
Rebased off of release/3.0b1. Original tip was 7814891.

jbudynk:
I've added tests for using logical operator `NOT` with comparison operator `!=`. I've also fixed a test that was failing due to a slightly unexpected warning message `FilterExpressionValidatorTest::testInvalidExpressionColumnDNE1()`

PR #327, opened 2016-08-11
Asynchronous SedEvent processing in the Visualizer

eholum:
Add an executor to linearly process SedEvents in the Visualizer.

Some of the VisualizerComponentTests are failing, but in general it seems pretty stable for 'normal' usage. At least I'm not seeing any clear evidence of concurrency problems with standard usage.


eholum:
This is now in a pretty solid place. Included many more changes for asynchronously handling of serialization in the SedModel. Ultimately I want to get serialization completely out of the SedModel and into iris-common or something, but that's down the road.

For now
+ Added a method for serializing an entire SED at once instead of one segment at a time (514 segments still takes ~3 minutes)
+ Removed async event processing from the VisualizerEventListener, put in error handling and retry function
+ Quite a bit of cleanup and simplification in the SedModel/DataStore.
+ Seds with more than 16 segments are now automatically plotted as a single layer.

PR #331, opened 2016-08-12
Additional Minor Bug Fixes

eholum:
Fixes:

+ Fit 3c066a with a broken powerlaw, then change the units and fit again: the fit is bogus.
+ the ranges window does not update automatically and you have to refresh it.
+ Clear all does not clear the ranges and the model plot.

Note this is based off of eholum-async, so that should be merged first.


olaurino:
Rebasing and fixing conflicts: original tip 24e2a7f.

Issue #332, opened 2016-09-16
Switching from linear to logarithmic scales on a Fixed plot with negative axes ranges


jbudynk:
An IllegalArgumentException occurs due to negative logarithmic ranges.

To see the error:
- Open the Visualizer. 
- Select "View --> Fixed Plot"
- Select "View --> Plot Type --> Linear"
- Zoom out so at least one of the axes has negative values in its range
- Switch back to log scale "View --> Plot Type --> Log"

The Plotter freezes, but one can work around this issue by clicking the "Reset" button to the Plotter area interactive again.

PR #333, opened 2016-10-05
Fixes issue 332 (plotting negative values on a fixed log plot)

jbudynk:
This PR addresses bug #332. 

When switching from a linear scale to a logarithmic scale on a fixed plot, the X and Y axis ranges are checked for negative values. If no negative axes ranges are found, the aspect is kept and used for the updated display. Otherwise, the display is reset to the full, default view of the SED. In other words, if _either axis_ has a minimum value of 0 or less, the whole plot is reset, even if the offending axis is to remain linear.

The check is done in StilPlotter.resetPlot().

A test which follows the steps reported in the #332 bug report is added to StilPlotterTest. Tests were also done by hand to check the code works as expected with various plot scale changes.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/StilPlotter.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/StilPlotter.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2015 Smithsonian Astrophysical Observatory
+ * Copyright (C) 2015,2016 Smithsonian Astrophysical Observatory
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


iris-visualizer/src/main/java/cfa/vo/iris/visualizer/plotter/StilPlotter.java
```diff
@@ -268,7 +268,14 @@ void resetPlot(boolean forceReset, boolean newPlot)
         // forcing a redraw
         if (fixed && !forceReset) {
             PlaneAspect existingAspect = getPlotPreferences().getAspect();
-            display.setAspect(existingAspect);
+            
+            // If switching any axis from linear to log scale, only keep the 
+            // original aspect if that axis has positive values. Otherwise, an 
+            // error will occur when trying to calc the log of a negative number.
+            if ((getPlotPreferences().getXlog() && existingAspect.getXMin() > 0)
```
jbudynk: Yes, you're right, this resets the linear scale for a fixed plot. Thanks for the catch.

jbudynk:
I've added a fix for @eholum 's note. I've tested the fix by hand, but I want to add a test before closing the PR.

jbudynk:
rebasing on master. Original tip: 1bf30fb

jbudynk:
I added the test for Erik's note in StilPlotterTest. PR is ready to merge, pending on review.

jbudynk:
rebasing on master. original tip was cab171b.

PR #335, opened 2016-10-12
Bug 334 error in legend label

### LINKED ISSUE #334: "_ERROR" is appended to each Segment label in the Plotter legend
On Iris the release/3.0b1 branch, if you load a NED SED and plot it, the resultant legend label shows the Segment's target name, followed by "_ERROR". For example, loading "m87" from NED results in the legend label "MESSIER 087_ERROR". It should be "MESSIER 087".

jbudynk:
Addresses #334 - "_ERROR" is appended to each Segment label in the Plotter legend

This change fixes #334 by setting the LayerModel's `leglabel` in VisualizerDataModel to the LayerModel's `suffix` (which by default is the segment StarTable's name). Like the metadata browser, the legend labels are incremented by 1, 2, 3, etc. for segments which share the same base table name. 

For example, if there are two segments with the target name "MESSIER 87", the first segment plotted has the label "MESSIER 87," while the second one is labeled "MESSIER 87 1". A third segment of the same name would be "MESSIER 87 2".

Before this fix, the LayerModel's `leglabel` property was always set to 'null'. By default, STILTS uses the `suffix` property as the Layer's label if the `leglabel` property isn't set. STILTS must have been picking up the error LayerModel's `suffix` for the some of the segment's labels. Setting the Layer's `leglabel` manually then forces STILTS to use the right label.

A test was added to StilPlotterTest to check that the `leglabel` property is not null and is set to the proper value.


jbudynk:
rebased on master. Original tip was fc29dc5

jbudynk:
Closing in preference to #339 (same changes, but merging into master branch).

Issue #336, opened 2016-10-14
clearing the model does not prevent the "Warning: SED changed" message from popping up when masking data


jbudynk:
To reproduce:
- create a new sed. Add a segment.
- open the fitting tool. add a model and click "Fit"
- clear the fit with "Clear All" from the fitting tool
- open the metadata browser, select a couple points, and click "Mask"

The warning pops up. I think if the fit is cleared, the warning should not pop up.

Also, the warning pops up multiple times. For example, after having masked a couple points, if I then click "Clear All" in the metadata browser, I'll get another warning.

jbudynk:
Along these lines, what if whenever this warning should pop up, we have the "fit status" bar update when the fit is no longer valid with the current SED? I think @olaurino mentioned this, too.

PR #338, opened 2016-12-05
Move "Add ranges" button and prevent unnecessary 'SED changed' warning

jbudynk:
Addresses:
#336 
ChandraCXC/iris-dev/issues/118

# Description

These changes are minor updates to the Fitting Tool:

- (ChandraCXC/iris-dev/issues/118) The "Add Ranges" button is moved to be above the "Fit" button and associated widgets.
- (#336) When a SED is fit, its model is cleared, and then changed (new segment, new masked points), a "SED changed" warning no longer appears. 

   The warning is called from `StilPlotter.validateModelFunctions()` if `SedModel.hasModelFunction` is True or if the SedModel's model version has changed. This PR sets `hasModelFunction` to False in `SedModel.clearFittingData()`.

# Testing

A test has been added to StilPlotter.testModelFunctionValidity() to check the fix to #336. Testing was done by hand to check that masking or otherwise editing the SED after the model was cleared did not cause the warning message to popup, and to check that the "Add Ranges" button still functions as expected.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/fitting/FitConfiguration.java
+++ b/iris-common/src/main/java/cfa/vo/iris/fitting/FitConfiguration.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright (C) 2016 Smithsonian Astrophysical Observatory
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


jbudynk:
Hold up... there's a failure in one of the tests due to something. Investigating...

jbudynk:
closing in preference to #340 (same changes, but based off of master branch).

PR #341, opened 2016-12-19
Iris dev #89 no warning save model

jbudynk:
# Description

This PR fixed ChandraCXC/iris-dev#89, where no warning message appeared if the user tries to overwrite an existing file. Now a warning message pops up asking the user if they wish to overwrite the file or not.

Also added to this PR is the removal of the 'Flux/Flux Density' spinner from the Plot window, as it is unused.

The `.gitignore` file is updated to exlcude *.DS_store files.

# Testing

Tests for overwriting and canceling the overwrite have been added to the functions `testSaveText()` and `testSaveJson()` in FittingToolComponentTest.java. Tests were also performed by hand to ensure the right confirmation message appeared when selecting an existing filename in the JFileChooser, and that all the buttons -- Yes, No, and the dispose 'X' button -- worked as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ target
 .idea
 *.iml
 nbactions.xml
+.DS_Store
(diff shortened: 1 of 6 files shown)


jbudynk:
Rebased on master. Original tip was ecbbab5.

Updated test in IrisFunctionalIT with new overwrite changes. I also removed the Flux/Flux Density spinner from the Plotter window, as it is unused.

PR #342, opened 2016-12-19
Make travis ci tests more likely to pass

olaurino:
There is a test consistently failing on travis with jdk8, but the test is OK when run (automatically or manually) on our development systems. We are ignoring the test for the time being.

This PR also removes an unnecessary loop waiting for menu to be ready, since that happens in other parts of the functional test now.

I am disabling the retry functionality, as it only takes longer for tests to fail. However, I increased the timeout for a test that was failing.

Unfortunately, there are still erratic failures, but the tests seem to pass more than 50% of the time now.


jbudynk:
A passing Travis build? Looks good to me ;)

But really, the code changes look good. I'm fine with the `testSetFittingRangesNoPlotter()` being skipped, as it's not that important. We should check by hand during 3.0 integration testing that this test passes.

OK to merge.

PR #343, opened 2016-12-19
Make sure clearAll button clears selected parameter, if any

olaurino:
This PR ensures that when clicking "Clear All" in the fitting tool, if a parameter is selected in the model view panel, then the parameter is cleared as well.


jbudynk:
Looks good. OK to merge.

PR #344, opened 2016-12-20
Fix irisdev 97 bulk operations

olaurino:
This PR introduces bulk operations for model parameters. All parameters in a model can be frozen/thawed by right clicking on the model in the models tree. All parameters in all models can be frozen/thawed by right clicking on the models tree root.

### DIFF
--- a/iris-common/src/main/java/cfa/vo/iris/fitting/FitConfiguration.java
+++ b/iris-common/src/main/java/cfa/vo/iris/fitting/FitConfiguration.java
@@ -43,6 +43,8 @@ public class FitConfiguration {
     public static final String PROP_EXPRESSION = "expression";
     public static final String PROP_CONFIDENCE = "confidence";
 
+    public static final String ROOT_MODELS_STRING = CompositeModelTreeModel.ROOT_STRING;
+
     private CompositeModel model;
     private List<FittingRange> fittingRanges;
     private Stat stat;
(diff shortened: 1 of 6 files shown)


jbudynk:
Works well. Model parameters are frozen/thawed as expected.

I think there should be a divider line between "Remove" and the "Freeze / Thaw" options, as a user might accidentally click the wrong menu item and remove their component by accident.

I like how it utilizes the tree structure for freezing/thawing. I might use it in the demo so I can freeze the parameters of one model while fitting another model.

@olaurino, what do you think of adding a menu item to the Fitting Tool for freeze/thaw all models? Before I read the directions, I didn't think to right-click to freeze/thaw parameters, but then again I'm used to the old Iris. I can live with or without this addition.

olaurino:
Ok, I'll add a line in the contextual menu and I'll try to add a drop-down menu... I was just hoping I could avoid firing up Netbeans ;)

olaurino:
As we discussed offline, adding the separator is trivial, but the drop-down menu less so. While the FreezeAll/ThawAll actions easily apply to the entire tree, having Freeze/Thaw bulk operations for all the parameters in a component is more complex, because it depends on which component is selected, but then the contextual menu is more user friendly, and a toolbar better suited than the drop-down. I pushed a commit introducing the separator, but we'll keep it simple for now and add an RFE for reviewing the UI later.

jbudynk:
The new changes look good. I give this my OK to merge.

PR #345, opened 2016-12-20
make sed extraction asynchronous and safer

olaurino:
I can't reproduce ChandraCXC/iris-dev#128. However, investigating the issue I found out the implementation of the extraction could be made safer and asynchronous. At the very least, if the issue can be reproduced on some systems, this PR should minimize it by a) not stopping the GUI thread and b) by creating and populating the SED first and adding it to the SedManager only when ready.

A couple of questions for @jbudynk on the UI. I didn't include a spinning wheel, do you think we need one? Also, I did not add a Stop button as the Stop functionality should be implemented, and I didn't look into what that would require. If you feel strongly we should add a Stop button I'll look into it, maybe after we have addressed the current outstanding issues.

### DIFF
--- a/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/MetadataBrowserMainView.java
+++ b/iris-visualizer/src/main/java/cfa/vo/iris/visualizer/metadata/MetadataBrowserMainView.java
@@ -16,13 +16,12 @@
 package cfa.vo.iris.visualizer.metadata;
 
 import cfa.vo.iris.gui.NarrowOptionPane;
+
+import java.util.concurrent.ExecutionException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import javax.swing.JOptionPane;
-import javax.swing.JPanel;
-import javax.swing.JScrollPane;
-import javax.swing.JTable;
+import javax.swing.*;
 
 import org.apache.commons.collections.CollectionUtils;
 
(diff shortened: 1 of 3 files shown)


olaurino:
On OSX I could not reproduce the issue, but I seemed to experience ChadraCXC/iris-dev#129 using the master branch. The issue seems to be resolved by this PR.

jbudynk:
I agree that this PR resolves ChandraCXC/iris-dev#129. I cannot reproduce the issue on this branch.

I don't think we need a spinning wheel; it's clear to see that the tool is 'thinking' while the OK message is popping up, as the "Extract" button remains highlighted blue. The data seems to load in fine, and I'm able to switch to other windows after clicking "OK."

However, it seems like the asynchronous work is only done on the second refresh of the data, and not the first one. For example, the GUI is frozen before the OK message pops up. Is this something we can fix before AAS?

As discussed offline, we might be able to remove the extra "refresh" that happens on updating SEDs, which can be clearly seen with these changes with large SEDs. However, this should be added to a different issue; it should not be addressed here.

jbudynk:
Ah woops, I might have understood the double refresh problem at first. The extra "refresh" I talk about in my previous comment must be the first SED refresh that occurs. So for the question I asked above -- can we do this asynchronously on the first refresh -- should be done in a different task.

olaurino:
I pushed a commit so the execution of preferences.assSed is sent to the EDT with invokeLater, which should prevent the GUI to freeze after the SED has been extracted and while it is added to the workspace.

jbudynk:
OK to merge.

PR #346, opened 2016-12-21
Handle index column sorting (and fix ChandraCXC/iris-dev#116)

olaurino:
This PR fixes ChandraCXC/iris-dev#116 by properly handling the index column.

There is an issue, though it pre-exists this PR: [this test](https://github.com/ChandraCXC/iris/blob/081f1a86078362748af0bacc2af985c3bffe73ff/iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/IrisStarJTableTest.java#L255) seems to suggest sorting order should be preserved when a segment is added to an SED. However, that does not seem the case when running Iris. This PR does not fix that issue, but the fix will hopefully be compatible with an actual fix.

### DIFF
--- a/iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/IrisStarJTableTest.java
+++ b/iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/IrisStarJTableTest.java
@@ -312,4 +312,36 @@ public class IrisStarJTableTest {
         assertEquals(true, table.getValueAt(2, 1));
         table.setSelectedStarTables(Arrays.asList(segTable));
     }
+
+    @Test
+    public void testRowSortErrorIndexCol() throws Exception {
+        IrisStarJTable table = new IrisStarJTable();
+        table.setSortBySpecValues(true);
+        table.setUsePlotterDataTables(true);
+
+        // Create Segment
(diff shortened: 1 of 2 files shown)


iris-visualizer/src/test/java/cfa/vo/iris/visualizer/metadata/IrisStarJTableTest.java
```diff
@@ -312,4 +312,36 @@ public void testRowSortErrorLastCol() throws Exception {
...
+        table.setSortBySpecValues(true);
+        table.setUsePlotterDataTables(true);
+
+        // Create Segment
+        IrisStarTable segTable = adapter.convertSegment(TestUtils.createSampleSegment(
+                new double[] {10,20,30}, new double[] {40,50,60}));
+        table.setSelectedStarTables(Arrays.asList(segTable));
+
+        // Sort by Original Flux Value (last column)
+        table.getRowSorter().setSortKeys(Arrays.asList(new RowSorter.SortKey(0, SortOrder.DESCENDING)));
```
jbudynk: The comment is misleading; the last column has index 5, right? I think it's meant to say "// Sort by Index (first column)"

jbudynk:
Looks good. I think this is fine to merge.

I noticed that the Index sorting remains when one switches between SEDs. I think it's fine for now,  but might need to change depending on what we decide on sorting permanence for the other columns.

PR #340, opened 2016-12-15
Move "Add ranges" button and prevent unnecessary 'SED changed' warning

jbudynk:
Addresses:
#336 
ChandraCXC/iris-dev/issues/118

# Description

These changes are minor updates to the Fitting Tool:

- (ChandraCXC/iris-dev/issues/118) The "Add Ranges" button is moved to be above the "Fit" button and associated widgets.
- (#336) When a SED is fit, its model is cleared, and then changed (new segment, new masked points), a "SED changed" warning no longer appears. 

   The warning is called from `StilPlotter.validateModelFunctions()` if `SedModel.hasModelFunction` is True or if the SedModel's model version has changed. This PR sets `hasModelFunction` to False in `SedModel.clearFittingData()`.

# Testing

A test has been added to StilPlotter.testModelFunctionValidity() to check the fix to #336. Testing was done by hand to check that masking or otherwise editing the SED after the model was cleared did not cause the warning message to popup, and to check that the "Add Ranges" button still functions as expected.

PR #347, opened 2016-12-21
Fix irisdev 131 update fit stats

olaurino:
This PR makes sure that the fit summary panel is updated when a SED is selected, not only after a fit.

A test was failing on Travis, so I added another timeout.

This PR also reverts back most of the changes in PR #340 that introduced issues in the Fitting Tool GUI


jbudynk:
Looks good for the most part.

While this wasn't mentioned in the original task, I noticed the model parameter fields don't update when switching between SEDs, unless one highlights a model parameter from the selected SED. 

There's still the GUI issue where the "Stop" button by the Confidence panel is hidden on OS X. Looks fine on Linux. Do you think this would be an OK PR to tag ChandraCXC/iris-dev#125 along with?

olaurino:
Let's address these issues after the dry run. I'll merge it and package Iris3.0b2 for now.

PR #356, opened 2017-02-21
Fix confidence timeout issue

olaurino:
Similarly to #352, this makes sure confidence calculations graciously time out.


olaurino:
Mistakenly issued against master

